Issue #35, opened 2022-06-08
Multi-tenancy: mutual exclusion for resources


shaneutt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem Statement

We want to support multi-tenancy for the operator, but currently we allow the same `controller-name` for `GatewayClass` to be present on two different deployments, and otherwise two separate deployments of the operator would both currently try to reconcile all related objects.

The purpose of this task is to ensure that a second operator which uses the exact same `GatewayClass` controller name will NOT start, so as to avoid two operators working on the same `Gateways`.

Additionally, only one operator should be responsible for any `DataPlane` or `ControlPlane` at a time, to ensure safety when deploying multiple copies of the Operator at different versions.

### Proposed Solution

- [ ] add `ControllerName` to the spec for each object we reconcile
- [ ] `ControllerName` for all objects should have automatic kubebuilder defaults
- [ ] the manager should be able to identify if a copy of the operator is already running with the given `ControllerName` and refuse to start if so

### Acceptance Criteria

- [ ] as a user I can change the controller name of the Gateway Operator in a manner consistent with class ingress class name for multi-tenancy
- [ ] as a user I don't have to worry about controller-names for my Gateway Operator unless I have multiple deployments: the defaults will be handled for me.
- [ ] if I deploy a second Gateway Operator using the exact same controller name as an existing one, the operator doesn't start and throws errors to indicate the problem
- [ ] if I have two Gateway Operators running on a cluster with separate controller names, they ignore eachothers resources and only operate on their own

Issue #22, opened 2022-06-07
GatewayClass Support


shaneutt:
### Problem Statement

For the POC a basic level of `Gateway` support was added but that just acts on _any_ `Gateway` resource.

The purpose of this task is to add `GatewayClass` support to the Gateway Operator.

### Acceptance Criteria

- [x] the operator provides a default `GatewayClass` controller name, which can be overridden by command line args AND env vars
- [x] when creating/updating `Gateways` the operator will only act on those configured for the `GatewayClass` objects which refer to our controller name as owner
- [x] creations and updates of `GatewayClass` objects trigger reconcilation for connected `Gateways`

### Additional Notes

- [x] TODO comments which linked to this issue are removed once resolved

PR #37, opened 2022-06-09
feat: GatewayClass support

shaneutt:

This adds support for `GatewayClass` objects so that we can clearly delineate which `Gateway` objects the controller supports. It also adds a configurable `controller-name` to specify that and help enable multi-tenancy (however do note, full multi-tenancy will also require #35).

**Which issue this PR fixes**

#22 

**PR Readiness Checklist**:

- [x] the `CHANGELOG.md` release notes have been updated


internal/utils/kubernetes/lists.go
```diff
@@ -9,12 +9,100 @@ import (
...
+	)
+	if err != nil {
+		return nil, err
+	}
+	selector := labels.NewSelector().Add(*requirement)
+
+	listOptions := &client.ListOptions{
+		LabelSelector: selector,
+	}
+	if gateway.Namespace != "" {
```
randmonkey, line 42: Since `gateway` is a namespaced resource, but using empty namespace is to list in all namespaces. I think it should return nothing when `gateway` does not have a namespace, because it may indicate an error. What was your expectation?

internal/utils/kubernetes/lists.go
```diff
@@ -9,12 +9,100 @@ import (
...
+		listOptions.Namespace = gateway.Namespace
```
randmonkey, line 85: same as the case before

randmonkey (changes requested):
Almost OK, but some minor suggestions and questions on listing and comments for functions

internal/utils/gateway/utils.go
```diff
@@ -0,0 +1,36 @@
...
+		if cond.Type == string(gatewayv1alpha2.GatewayConditionScheduled) &&
+			cond.Reason == string(gatewayv1alpha2.GatewayReasonScheduled) &&
+			cond.Status == metav1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
+
+// IsGatewayScheduled indicates whether or not the provided Gateway object was
```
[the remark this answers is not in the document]
shaneutt: Good catch! fixed in 1540197b022489f96ad005047be9fcb9cdfc6954

internal/utils/kubernetes/lists.go
shaneutt: Sure that's fair, changed in bcfe3da78648b4206a627fe7f3b2d0db7e755ba8 :+1:

internal/utils/kubernetes/lists.go
shaneutt: Changed in bcfe3da78648b4206a627fe7f3b2d0db7e755ba8 :+1:

randmonkey (approved):

Issue #13, opened 2022-06-06
Validation: Only Allow DBLESS DataPlanes


shaneutt:
### Problem Statement

We're not intending to support anything but DBLESS mode for the Kong Gateway via `DataPlane` right now. The purpose of this task is to add validation to explicitly disallow this.

### Additional Notes

We should create a validation library (e.g. `internal/validation` similar to KIC) which does the validation in one place, and call it from both the webhook and controller to re-use code.

### Acceptance Criteria

- [ ] webhook is added to our deployment, and automatically deployed in all test suites (e.g. `integration`, `conformance`, `e2e`, e.t.c.)
- [ ] webhook validation prohibits anything but DBLESS mode for the `DataPlane` as we don't (currently) support this.
- [ ] controller validation prohibits anything but DBLESS mode for the `DataPlane` as we don't (currently) support this.

PR #39, opened 2022-06-13
feat(validation) add validation for dataplane

### LINKED ISSUE #38: Need a certificate for webhook server
## Problem Statement
Need a method to provide certificate for webhook server, for example use `certManager` to provide a certificate.

## Additional Notes

Working on #13 and blocked on the issue that running a webhook server when testing validation webhook server. We need a method to provide a certificate for webhook server

## Acceptance Criteria
- [ ] could run webhook server in operator without additional work on providing certificates.

randmonkey:

Added validator interface and dataplane validation for dbMode in Reconciler and Webhook. Added event recorder to show the validation result during reconciling in `kubectl describe`.

**Which issue this PR fixes** 
#13, also would fix #38

Work in progress, still working on adding a certificate(probably by `certManager`) for webhook server

**PR Readiness Checklist**:


controllers/dataplane_controller.go
```diff
@@ -85,6 +91,14 @@ func (r *DataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 		return ctrl.Result{}, nil // no need to requeue, the update will trigger.
 	}
+	// validate dataplane
+	v := dataplanevalidation.NewValidator(r.Client)
+	err = v.Validate(dataplane)
```
shaneutt, line 96: as we only need to use the validator once and it's constructor doesn't return an error, are you opposed to daisy-chaining the call to save a line?

```suggestion
	err = dataplanevalidation.NewValidator(r.Client).Validate(dataplane)
```

controllers/dataplane_controller.go
```diff
@@ -85,6 +91,14 @@ func (r *DataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	if err != nil {
+		debug(log, "failed to validate dataplane: error "+err.Error(), dataplane)
```
shaneutt, line 98: across KIC and other codebases of ours, it's pretty common to just put the error itself after the colon, if that sounds reasonable to you:

```suggestion
		debug(log, "failed to validate dataplane: "+err.Error(), dataplane)
```

controllers/dataplane_controller.go
```diff
@@ -85,6 +91,14 @@ func (r *DataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		r.eventRecorder.Event(dataplane, "Warning", "ValidationFailed", err.Error())
```
shaneutt, line 99: :+1: I'm glad that you took the time to add an event recorder and emit events about this. This reminds me of https://github.com/Kong/kubernetes-ingress-controller/issues/2195 where we are in need of something like this for the KIC.

controllers/dataplane_controller.go
```diff
@@ -85,6 +91,14 @@ func (r *DataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		return ctrl.Result{Requeue: false}, nil
```
shaneutt, line 100: So we have a log entry AND an event for this, and that's good but we should keep in mind that both logs and events are potentially transient. We should update the status of the object as well to reflect the problem: we have the `DataPlaneConditionTypeProvisioned` status condition type already to indicate whether the `DataPlane` object has been provisioned. I think it would be fair to add a new `reason` such as `DataPlaneConditionReasonFailedValidation` and when we get to this point, ensure that the status conditions include a provisioned condition with that reason and `False`, plus a helpful message that indicates the exact validation failure.

internal/admission/server.go
```diff
@@ -0,0 +1,154 @@
...
+
+	operatorv1alpha1 "github.com/kong/gateway-operator/api/v1alpha1"
+)
+
+var (
+	scheme = runtime.NewScheme()
+	codecs = serializer.NewCodecFactory(scheme)
+)
+
+func NewWebhookServerFromManager(mgr ctrl.Manager) *webhook.Server {
```
shaneutt, line 25: lint: public symbols need godoc comments

internal/admission/server.go
```diff
@@ -0,0 +1,154 @@
...
+	hookServer := mgr.GetWebhookServer()
+	handler := NewRequestHandler(logrus.New())
+	hookServer.Register("/validate", handler)
+	return hookServer
+}
+
+type Validator interface {
```
shaneutt, line 32: lint: public symbols need godoc comments

internal/admission/server.go
```diff
@@ -0,0 +1,154 @@
...
+	ValidateControlPlane(context.Context, operatorv1alpha1.ControlPlane) error
+	ValidateDataPlane(context.Context, operatorv1alpha1.DataPlane) error
+}
+
+type RequestHandler struct {
```
shaneutt, line 37: lint: public symbols need godoc comments

internal/admission/server.go
```diff
@@ -0,0 +1,154 @@
...
+	// Validator validates the entities that the k8s API-server asks
+	// it the server to validate.
+	Validator Validator
+
+	Logger logrus.FieldLogger
+}
+
+func NewRequestHandler(logger logrus.FieldLogger) *RequestHandler {
```
shaneutt, line 45: lint: public symbols need godoc comments

internal/admission/server.go
```diff
@@ -0,0 +1,154 @@
...
+	return &RequestHandler{
+		Validator: &validator{},
+		Logger:    logger.WithField("a", "b"),
```
shaneutt, line 48: Should we update these fields?

internal/admission/server.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+func (h *RequestHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
shaneutt, line 52: lint: public symbols need godoc comments

internal/admission/server.go
```diff
@@ -0,0 +1,154 @@
...
+			_, _, err := deserializer.Decode(req.Object.Raw, nil, &controlPlane)
+			if err != nil {
+				return nil, err
+			}
+			err = h.Validator.ValidateControlPlane(ctx, controlPlane)
+			if err != nil {
+				ok = false
+				msg = err.Error()
+			}
+		}
```
shaneutt, line 129: I see you're taking the time to just scaffold out validation for the `ControlPlane` as well while we're already here, :+1:

internal/consts/consts.go
```diff
@@ -55,3 +55,10 @@ const (
 	// all other attempts to dynamically decide the default image fail.
 	DefaultControlPlaneImage = DefaultControlPlaneBaseImage + ":" + DefaultControlPlaneTag
 )
+
+// -----------------------------------------------------------------------------
+// Consts - Environment Variable Names
+// -----------------------------------------------------------------------------
```
shaneutt, line 61:
```suggestion
// -----------------------------------------------------------------------------

```

internal/consts/consts.go
```diff
@@ -55,3 +55,10 @@ const (
...
+const (
+	EnvVarKongDatabase = "KONG_DATABASE"
```
shaneutt, line 63: lint: public symbols need godoc comments

internal/manager/run.go
```diff
@@ -27,10 +27,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
```
shaneutt, line 30: lint: I think this blank space is making the linter mad
```suggestion
```

internal/manager/run.go
```diff
@@ -121,6 +123,19 @@ func Run(cfg Config) error {
 		return fmt.Errorf("unable to set up ready check: %w", err)
 	}
 
+	tlsCertPath := "/tmp/k8s-webhook-server/serving-certs/tls.crt"
+	tlsKeyPath := "/tmp/k8s-webhook-server/serving-certs/tls.key"
```
shaneutt, line 127: I take it since this PR is in draft still, this is a temporary hardcoded location and we'll add a dynamic method to import these soon?

internal/validation/dataplane/validation.go
```diff
@@ -0,0 +1,88 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	operatorv1alpha1 "github.com/kong/gateway-operator/api/v1alpha1"
+	"github.com/kong/gateway-operator/internal/consts"
+)
+
+type Validator struct {
```
shaneutt, line 16: lint: this and many of the public symbols in this file need godoc comments

internal/validation/dataplane/validation.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (v *Validator) Validate(dataplane *operatorv1alpha1.DataPlane) error {
+	err := v.ValidateDeployOptions(dataplane.Namespace, &dataplane.Spec.DeploymentOptions)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (v *Validator) ValidateDeployOptions(namespace string, opts *operatorv1alpha1.DeploymentOptions) error {
```
shaneutt, line 32: I appreciate that you did this validation against `DeploymentOptions`, that's going to work to our advantage because the upcoming `GatewayConfiguration` API is going to have to use this validator as well :+1:

internal/validation/dataplane/validation.go
```diff
@@ -0,0 +1,88 @@
...
+			} else if envVar.ValueFrom != nil {
+				// value is empty,get from ValueFrom from configmap/secret.
+				if envVar.ValueFrom.ConfigMapKeyRef != nil {
+					cmKeyRef := envVar.ValueFrom.ConfigMapKeyRef
+					cm := &corev1.ConfigMap{}
+					namespacedName := k8stypes.NamespacedName{Namespace: namespace, Name: cmKeyRef.Name}
+					err := v.c.Get(context.Background(), namespacedName, cm)
+					if err != nil {
+						return fmt.Errorf("failed to get configMap %s in configMapKeyRef, error %w", cmKeyRef.Name, err)
+					}
```
shaneutt, line 53: I agree with the choice to throw an error here :+1: 

It's a bit against the eventual consistency grain of Kubernetes APIs, but if we _didn't_ throw an error here then we would either a) have to validate `ConfigMaps` on creation referenced against `DataPlanes` which reference them (which is a nightmare) or b) we would have to do an extra level of validation during reconcilation once we finally have the values from the `ConfigMaps` pulled, which I think is potentially rough UX.

That said, because this goes against the grain a bit I think it's worth adding a comment here so that future code readers can quickly understand the nuance introduced here:

```suggestion
					// technically it goes slightly against eventual-consistency to throw an error here,
					// but the alternative is that we would need to validate ALL ConfigMaps on create
					// and do relational mapping to DataPlane resources to validate that they aren't
					// going to introduce a new violation, or we would have to do an additional level
					// of validation that could only run during reconcilation.
					err := v.c.Get(context.Background(), namespacedName, cm)
					if err != nil {
						return fmt.Errorf("can't validate referenced configMap %s: %w", cmKeyRef.Name, err)
					}
```

Also as you'll see I suggested an update to the user-facing error message: for something a bit out of the norm like this I think the language we provide to the API caller is very important. Let me know what you think of that suggestion there.

As an aside: ultimately doing this kind of cross-reference validation is a tricky balancing act: upstream might push us _not_ to do it and in a general sense they wouldn't be wrong, but the reality is that a clear error upfront can go a long way for user experience. I think this change strikes a good balance, and the other good thing is it's easy to remove this later if it somehow was causing problems.

internal/validation/dataplane/validation.go
```diff
@@ -0,0 +1,88 @@
...
+				}
+
+				if envVar.ValueFrom.SecretKeyRef != nil {
+					secretRef := envVar.ValueFrom.SecretKeyRef
+					secret := &corev1.Secret{}
+					namespacedName := k8stypes.NamespacedName{Namespace: namespace, Name: secretRef.Name}
+					err := v.c.Get(context.Background(), namespacedName, secret)
+					if err != nil {
+						return fmt.Errorf("failed to get secret %s in secretRef, error %w", secretRef.Name, err)
```
shaneutt, line 66: Similar to the above comment, I agree with the approach I just think there's enough nuance here that it's worth adding some comments and maybe adjusting the error message for considerations around end-user clarity.

internal/validation/dataplane/validation.go
```diff
@@ -0,0 +1,88 @@
...
+							dbMode = string(decoded)
+							dbModeFound = true
+						}
+					}
+				}
+			}
+		}
+	}
+
+	// TODO: if dbMode not found in envVar, search for it in EnvVarFrom.
```
shaneutt, line 80: Don't forget to clean up this TODO :+1:

internal/validation/dataplane/validation_test.go
```diff
@@ -0,0 +1,187 @@
...
+			c: b.Build(),
+		}
+		err := v.Validate(tc.dataplane)
+		if !tc.hasError {
+			require.NoErrorf(t, err, tc.msg)
+		} else {
+			require.ErrorContainsf(t, err, tc.errMsg, tc.msg)
+		}
+	}
+}
```
shaneutt, line 187: Just a note that unless I missed it, it doesn't look like we've got this webhook completely hooked up in the kustomization `config/` configurations. Let's make sure we do that, and we add a basic webhook test to the E2E tests before we merge (e.g. the E2E tests just deploy via kustomize, ensure that the webhook is created and working properly by adding a call to try and create an invalid `DataPlane` to an existing test).

shaneutt (changes requested):
Early review, as requested. Things are looking really good! I have a few comments for your considerations. :vulcan_salute:

internal/validation/dataplane/validation.go
jrsmroz: I actually find this deeply nested code hard to follow.

shaneutt: > I actually find deeply nested conditional code hard to follow.

@jrsmroz It seems we should phrase that feedback an action: @randmonkey in addition to some comments and some updated error messages would you mind please refactoring this into more smaller tightly purposes functions/methods please?

internal/admission/server.go
randmonkey: Sorry, this is just a temporary thing,

internal/consts/consts.go
randmonkey: empty line here?

internal/manager/run.go
randmonkey: I think we would define the default paths as constants.

controllers/dataplane_controller.go
randmonkey: Yep, since there are currently no error returned from `NewValidator`, it is better to use a chained calling.

controllers/dataplane_controller.go
randmonkey: Maybe we could unify the logging format we are using. there are already 2 formats of logs:
```
{"level":"info","ts":1655179827.2874749,"logger":"controller-runtime.metrics","msg":"Metrics server is starting to listen","addr":"127.0.0.1:8080"}
```
from `zap`
```
I0614 04:10:27.288748       1 leaderelection.go:248] attempting to acquire leader lease kong-system/a7feedc84.konghq.com...
```
from `klog`(logging packages inside kubernetes)
and importing `logrus` brings another format:
```
time="2022-06-14T04:15:18Z" level=error msg="failed to run validation" component=validation-server
``` 
How about unifying to `zap`?

shaneutt: It still occurs to me that we would want a status condition for this of type `DataPlaneConditionTypeProvisioned` and `metav1.ConditionFalse` and an explanation therein.

internal/admission/server.go
shaneutt: ok :+1:

internal/manager/run.go
shaneutt: That's fine, and I know this is still draft too. Let's try to match how KIC does this to some degree when we get back around to it. :+1:

controllers/dataplane_controller.go
jrsmroz: +1 for unifying logging format.

internal/validation/dataplane/validation.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	return dbMode, dbModeFound, nil
+}
+
+func (v *Validator) getValueFromConfigMapKeyRef(namespace string, cmKeyRef *corev1.ConfigMapKeySelector) (string, bool, error) {
+	cm := &corev1.ConfigMap{}
+	namespacedName := k8stypes.NamespacedName{Namespace: namespace, Name: cmKeyRef.Name}
+	err := v.c.Get(context.Background(), namespacedName, cm)
+	if err != nil {
+		return "", false, fmt.Errorf("failed to get configMap %s in configMapKeyRef, error %w", cmKeyRef.Name, err)
```
shaneutt, line 131: commonly we use a colon to separate the messaging about the error from the error itself in a wrapped error as this usually ends up being a pretty clear delineation:

```suggestion
		return "", false, fmt.Errorf("failed to get configMap %s in configMapKeyRef: %w", cmKeyRef.Name, err)
```

internal/validation/dataplane/validation.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	return "", false, nil
+}
+
+func (v *Validator) getValueFromSecretRef(namespace string, secretKeyRef *corev1.SecretKeySelector) (string, bool, error) {
+	secret := &corev1.Secret{}
+	namespacedName := k8stypes.NamespacedName{Namespace: namespace, Name: secretKeyRef.Name}
+	err := v.c.Get(context.Background(), namespacedName, secret)
+	if err != nil {
+		return "", false, fmt.Errorf("failed to get secret %s in secretRef, error %w", secretKeyRef.Name, err)
```
shaneutt, line 144:
```suggestion
		return "", false, fmt.Errorf("failed to get secret %s in secretRef: %w", secretKeyRef.Name, err)
```

internal/validation/dataplane/validation.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	return "", false, nil
+}
+
+func (v *Validator) getDBModeFromConfigMapRef(namespace string, prefix string, cmRef *corev1.ConfigMapEnvSource) (string, bool, error) {
+	cm := &corev1.ConfigMap{}
+	namespacedName := k8stypes.NamespacedName{Namespace: namespace, Name: cmRef.Name}
+	err := v.c.Get(context.Background(), namespacedName, cm)
+	if err != nil {
+		return "", false, fmt.Errorf("failed to get configMap %s in configMapRef, error %w", cmRef.Name, err)
```
shaneutt, line 160:
```suggestion
		return "", false, fmt.Errorf("failed to get configMap %s in configMapRef: %w", cmRef.Name, err)
```

internal/validation/dataplane/validation.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	if cm.Data == nil {
+		return "", false, nil
+	}
+
+	// find key in Data that would become `KONG_DATABASE` after concatated to the prefix.
```
shaneutt, line 167:
```suggestion
	// find the key in the Data that would become `KONG_DATABASE` after concatenation with the prefix.
```

test/integration/suite_test.go
```diff
@@ -42,6 +44,9 @@ const gatewayAPIsCRDs = "https://github.com/kubernetes-sigs/gateway-api.git/conf
 var (
 	existingClusterName  = os.Getenv("KONG_TEST_CLUSTER")
 	controllerManagerOut = os.Getenv("KONG_CONTROLLER_OUT")
+	runWebhookTests      = false
+	webhookCertDir       = ""
+	webhookServerIP      = ""
```
shaneutt, line 49: Why not pull this directly from the env?

```suggestion
	webhookServerIP      = os.Getenv("GATEWAY_OPERATOR_WEBHOOK_IP")
```

test/integration/suite_test.go
```diff
@@ -121,8 +136,27 @@ func TestMain(m *testing.M) {
...
 
+	if runWebhookTests {
+		fmt.Println("INFO: creating certificates for running webhook tests")
+		dir, err := os.MkdirTemp(os.TempDir(), "gateway-operator-webhook-certs")
+		exitOnErr(err)
+		webhookCertDir = dir
+		fmt.Println("INFO: creating certificates in directory", webhookCertDir)
+		cmd := exec.CommandContext(ctx, "../../hack/generate-certificates-openssl.sh", webhookCertDir, webhookServerIP)
+		exitOnErr(cmd.Run())
+	}
```
shaneutt, line 147: Does this need to be separated from the code above on line 88? It seems like it's all related so perhaps we can create a separate function that's responsible for all the webhook and webhook-cert related steps?

test/integration/suite_test.go
```diff
@@ -121,8 +136,27 @@ func TestMain(m *testing.M) {
...
+	startControllerManager()
+
+	if runWebhookTests {
+		fmt.Println("INFO: creating a validating webhook and waiting for it to start")
+		err = createValidatingWebhook(ctx, k8sClient, fmt.Sprintf("https://%s:9443/validate", webhookServerIP),
+			webhookCertDir+"/ca.crt")
+		exitOnErr(err)
+		// wait for webhook to be ready
+		exitOnErr(waitForWebhook(ctx, webhookServerIP, 9443))
+	}
```
shaneutt, line 159: Similar to above comments, it seems if we just combined this with other webhook related setup it would help improve clarity in the suite, and I'm also not seeing the separation as necessary?

test/integration/util_test.go
```diff
@@ -33,3 +40,77 @@ func setup(t *testing.T) (*corev1.Namespace, *clusters.Cleaner) {
...
+					AdmissionReviewVersions: []string{"v1", "v1beta1"},
+				},
+			},
+		},
+		metav1.CreateOptions{})
+	return err
+}
+
+// returns the first found non-loopback IPv4 ip of local interfaces.
+func getFirstNonLoopbackIP() (string, error) {
```
shaneutt, line 90: Given that this is a testing environment, any reason we can simply listen on `0.0.0.0` here to always ensure we're listening on the outside routable network even if the networking configuration of the testing environment were to be changed down the road?

shaneutt (changes requested):
Looking good, I do have a few comments let me know your thoughts.

controllers/dataplane_controller.go
```diff
@@ -67,6 +87,18 @@ func (r *DataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	err = dataplanevalidation.NewValidator(r.Client).Validate(dataplane)
+	if err != nil {
+		debug(log, "failed to validate dataplane: "+err.Error(), dataplane)
+		r.eventRecorder.Event(dataplane, "Warning", "ValidationFailed", err.Error())
+		markErr := r.ensureDataPlaneIsMarkedNotProvisioned(ctx, dataplane,
+			DataPlaneConditionValidationFailed, err.Error())
+		if markErr != nil {
+			log.Error(err, "failed to mark dataplane as not provisioned")
+		}
+		return ctrl.Result{Requeue: false}, nil
```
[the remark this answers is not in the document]
randmonkey: updated.

controllers/dataplane_controller.go
```diff
@@ -67,6 +87,18 @@ func (r *DataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 		return ctrl.Result{}, nil // no need to requeue, the update will trigger.
 	}
+	// validate dataplane
+	err = dataplanevalidation.NewValidator(r.Client).Validate(dataplane)
+	if err != nil {
+		debug(log, "failed to validate dataplane: "+err.Error(), dataplane)
```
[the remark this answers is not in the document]
randmonkey: updated to `info()` here.

controllers/dataplane_controller.go
```diff
@@ -20,17 +22,35 @@ import (
...
+//+kubebuilder:rbac:groups=gateway-operator.konghq.com,resources=dataplanes/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=gateway-operator.konghq.com,resources=dataplanes/finalizers,verbs=update
+//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=create;get;list;watch;update;patch
+//+kubebuilder:rbac:groups=apps,resources=deployments/status,verbs=get
+//+kubebuilder:rbac:groups=core,resources=services,verbs=create;get;list;watch;update;patch
+//+kubebuilder:rbac:groups=core,resources=services/status,verbs=get
+//+kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch
+//+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+
```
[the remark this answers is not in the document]
randmonkey: removed the annotations here and updated in `controllers/dataplane_controller_rbac.go` .

controllers/dataplane_controller_reconciler_utils.go
```diff
@@ -63,6 +63,27 @@ func (r *DataPlaneReconciler) ensureDataPlaneIsMarkedProvisioned(
...
+			condition.ObservedGeneration = dataplane.Generation
+			condition.LastTransitionTime = metav1.Now()
+		}
+		updatedConditions = append(updatedConditions, condition)
+	}
+
+	dataplane.Status.Conditions = updatedConditions
+	return r.Status().Update(ctx, dataplane)
+}
+
```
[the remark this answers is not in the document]
randmonkey: Fixed. The current implementation goes as follow:
- if the condition with `DataPlaneConditionTypeProvisioned` exists and is not the same as expected, change it to the expected condition and update the status.
- if the condition with `DataPlaneConditionTypeProvisioned` does not exist, append our expected condition, and update the status.

internal/validation/dataplane/validation.go
randmonkey: fixed.

internal/validation/dataplane/validation.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// ValidateDeployOptions validates the DeploymentOptions field of DataPlane object.
+func (v *Validator) ValidateDeployOptions(namespace string, opts *operatorv1alpha1.DeploymentOptions) error {
+
+	// validata db mode.
```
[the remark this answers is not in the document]
randmonkey: fixed.

internal/validation/dataplane/validation.go
randmonkey: updated, and the same for the following.

test/integration/suite_test.go
randmonkey: updated

test/integration/suite_test.go
randmonkey: fixed for this. put the getting IP and generating certificates into one block.

test/integration/suite_test.go
randmonkey: I think the separation is necessary.  generating certificates should be done before we run controller manager, and waiting for webhook server to be available should be happen after controller manager start.

test/integration/util_test.go
randmonkey: Yes, our service do listen on `0.0.0.0`. But getting a specific IP is still necessary. The IP  is used for generating certificates and for clients to access. We need to specify an IP when we generate the certificate requests, and give an IP for clients to access.

randmonkey:
`make lint` complains 
```
INFO fixer took 0s with no stages                 
controllers/utils.go:26:92: directive `//nolint:deadcode,unused //FIXME` is unused for linter "deadcode" (nolintlint)
```
I searched for this error message and tried to fix it, but no effect. @shaneutt do you know how to fix the lint issue?

pmalek:
> `make lint` complains
> 
> ```
> INFO fixer took 0s with no stages                 
> controllers/utils.go:26:92: directive `//nolint:deadcode,unused //FIXME` is unused for linter "deadcode" (nolintlint)
> ```
> 
> I searched for this error message and tried to fix it, but no effect. @shaneutt do you know how to fix the lint issue?

@randmonkey remove `deadcode` from this `nolint` comment. This code is in fact not "dead": this **is** being used in here https://github.com/randmonkey/gateway-operator/blob/issue13_validation/controllers/dataplane_controller.go#L82

controllers/dataplane_controller_reconciler_utils.go
```diff
@@ -63,6 +63,55 @@ func (r *DataPlaneReconciler) ensureDataPlaneIsMarkedProvisioned(
 	return r.Status().Update(ctx, dataplane)
 }
 
+// isSameCondition returns true if two `metav1.Condition` indicates the same condition.
+func isSameCondition(condition1, condition2 metav1.Condition) bool {
+	return condition1.Type == condition2.Type &&
+		condition1.Status == condition2.Status &&
+		condition1.Reason == condition2.Reason &&
+		condition1.Message == condition2.Message
+	// TODO: should we compare `ObservedGeneration` here?
```
shaneutt, line 72: We should resolve this TODO, and I think the answer to the question is ultimately "no", it doesn't seem like what we're doing here is actually checking if it's the _exact_ same condition but instead if the non-database related fields are the same. We should update the function docs and maybe the name to reflect that.

shaneutt (changes requested):
I asked for this previously but the conversation was marked resolved without resolution: we need to add a test that ensures the webhook is functioning properly when the operator is deployed via kustomize (e.g. the contents in `config/`). I want to repeat my suggestion that we should add an E2E test that deploys the operator via `kustomize` and verifies that the webhook is functioning there, as this will help us prevent regressions. The KIC has some examples of deploying via kustomize for E2E tests which may be helpful.

randmonkey:
> I asked for this previously but the conversation was marked resolved without resolution: we need to add a test that ensures the webhook is functioning properly when the operator is deployed via kustomize (e.g. the contents in `config/`). I want to repeat my suggestion that we should add an E2E test that deploys the operator via `kustomize` and verifies that the webhook is functioning there, as this will help us prevent regressions. The KIC has some examples of deploying via kustomize for E2E tests which may be helpful. 
 Added e2e  tests in the latest PR. Because the operator image has not updated, the e2e fails now. If image `richardyi/kong-gateway-operator:20220713-1` is used in e2e tests, the test would be passed. The further things to do about e2e is to build and push a new image before running tests.

shaneutt (changes requested):

test/e2e/webhook_test.go
```diff
@@ -0,0 +1,73 @@
...
+			if tc.errMsg == "" {
+				require.NoErrorf(t, err, "test case %s: should not return error when creating dataplane", tc.name)
+			} else {
+				require.Error(t, err, "test case %s: should return error", tc.name)
+				require.Containsf(t, err.Error(), tc.errMsg, "test case %s: error message should contain expected content", tc.name)
+			}
+
+		})
+	}
+}
```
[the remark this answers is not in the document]
randmonkey: added steps of building image and loading image in e2e tests.

.github/workflows/tests.yaml
```diff
@@ -114,6 +114,14 @@ jobs:
...
+      env:
+        IMG: gateway-operator
+        TAG: e2e-${{ github.sha }}
+      run: make docker.build
+    
     - name: run e2e tests
-      run: make test.e2e
+      run: make test.e2e.github
+      env:
+        KONG_TEST_GATEWAY_OPERATOR_IMAGE_LOAD: gateway-operator:e2e-${{ github.sha }}
```
shaneutt, line 127: :+1: thanks for adding the image build and load

config/certificate/issuer.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: selfsigned-issuer
+  namespace: kong-system
+spec:
+  selfSigned: {}
```
shaneutt, line 7: We can't rely on cert-manager to be available in all cases, so adding this will potentially cause deployment to fail. We need to manage certificates for the webhook in a fully native manner.

shaneutt (changes requested):
I think we're getting _very_ close with this PR but I do think we have at least one remaining blocker: the way we're handling certificates for the webhook isn't portable.

I would suggest that we take a look at Gateway APIs for an example of how to manage webhook and certs in a native manner:

https://github.com/kubernetes-sigs/gateway-api/blob/main/config/webhook/certificate_config.yaml

config/certificate/issuer.yaml
randmonkey: removed those certmanager` configurations and used the same way to generate certificates as in Gateway API project.

Makefile
```diff
@@ -235,6 +235,11 @@ test.integration:
 test.e2e:
 	GOFLAGS="-tags=e2e_tests" go test -race -v ./test/e2e/...
 
+# the same as e2e, but used for github actions because the github workflow for e2e would modify contents of  manifests in `config`.
+test.e2e.github:
+	GOFLAGS="-tags=e2e_tests" go test -race -v ./test/e2e/...
+
```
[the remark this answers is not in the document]
randmonkey: fixed.

test/e2e/utils_test.go
```diff
@@ -0,0 +1,77 @@
...
+		image = imageOverride
+	}
+
+	if image == "" {
+		fmt.Println("INFO: use default image")
+		return nil
+	}
+
+	// TODO: deal with image names in format <host>:<port>/<repo>/<name>:[tag]
+	// e.g localhost:32000/kong/gateway-operator:xxx
```
shaneutt, line 50: Let's add an link a github issue to this TODO so that it's not lost

shaneutt (dismissed):
Looking good. Ultimately this PR has gotten quite large and so I imagine that there's still quite a few more things that we could touch up, but since we're still in a nice cocooned early development stage I think it's reasonable to take this value in its current state and iterate on it as we run into problems so :heavy_check_mark: let's get this merged.

Before merging however, please resolve all `TODO` comments: let's not leave errant `TODO` comments. They either should be resolved in the same PR _or_ at least create a follow up issue and link it in the comment so that they don't get out of band.

test/e2e/utils_test.go
```diff
@@ -0,0 +1,77 @@
...
+	// append image contents to replace image
+	fmt.Println("INFO: replacing image in kustomization file")
+	appendImageKustomizationContents := []byte(fmt.Sprintf(gatewayOperatorImageKustomizationContents, imageName, imageTag))
+	newBuf := append(buf, appendImageKustomizationContents...)
+	err = os.WriteFile("../../config/default/kustomization.yaml", newBuf, os.ModeAppend)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
pmalek, line 76:
```suggestion
	return os.WriteFile("../../config/default/kustomization.yaml", newBuf, os.ModeAppend)
```

Apart from the suggestion I'm kind of worried about the implicit dependency on this file. Is this being used from elsewhere? 

Also, not trying to pick on the name here but `setOperatorImage` doesn't really say much about the side effect I've mentioned about. There's also no comment on it which might exacerbate the surprise effect.

Maybe we could write that new kustomization file to a temp file and pass its path around and then ensure at the end of the tests that we clean it up? This doesn't have to happen in this PR but I'd really like to see this change given what I've described above.

test/e2e/utils_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/kong/kubernetes-testing-framework/pkg/clusters"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// namespace create a namespace for a test case, and clean it up after the test finishes.
```
pmalek, line 17:
```suggestion
// namespace creates a namespace for a test case and cleans it up after the test finishes.
```

Actually, this does not clean this up but it returns a `func` that can clean it up.

What you could do here is use `t.Cleanup()` to register the cleanup like so:

```go
func namespaceForCluster(t *testing.T) *corev1.Namespace {
	namespaceForTestCase, err := clusters.GenerateNamespace(ctx, env.Cluster(), t.Name())
	require.NoError(t, err) // this or return an error but NoError is "equivalent" to exitOnErr(err)

	t.Cleanup(func(){
		assert.NoError(t, clusters.CleanupGeneratedResources(ctx, env.Cluster(), t.Name()))
	})

	return namespaceForTestCase
}
```

test/e2e/utils_test.go
shaneutt: Good point. I agree that it would be nice if we could make a tempfile for this purpose instead of overwriting files in the repository before we ship this, as inevitably someone is going to accidentally commit those changes at some point.

test/e2e/utils_test.go
randmonkey: added issue #88.

test/e2e/utils_test.go
randmonkey: fixed. For the second question, I am looking at kustomize to find out how to do it.

test/e2e/utils_test.go
randmonkey: fixed comments.

Makefile
```diff
@@ -250,6 +250,10 @@ test.integration:
 test.e2e:
 	GOFLAGS="-tags=e2e_tests" go test -race -v ./test/e2e/...
 
+# the same as e2e, but used for github actions because the github workflow for e2e would modify contents of manifests in `config`.
```
pmalek, line 253: I don't understand this comment given that those 2 targets are exactly the same. Do you mind explaining further?

config/admission_webhook/certificate_config.yaml
```diff
@@ -0,0 +1,144 @@
...
+          env:
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+      restartPolicy: OnFailure
+      serviceAccountName: gateway-operator-admission
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 2000
```
pmalek, line 144: EOL

config/admission_webhook/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
+- certificate_config.yaml
+- service_validating_webhook.yaml
+- validating_webhook.yaml
```
pmalek, line 4: EOL

config/default/manager_webhook_certificates_patch.yaml
```diff
@@ -0,0 +1,26 @@
...
+            path: tls.crt
+          - key: key
+            path: tls.key
+      containers:
+      - name: manager
+        volumeMounts:
+        - name: webhook-certificates
+          mountPath: /tmp/k8s-webhook-server/serving-certs
+          readOnly: true
+
```
pmalek, line 26: EOL

internal/validation/dataplane/validation.go
```diff
@@ -0,0 +1,194 @@
...
+	return &Validator{c: c}
+}
+
+// Validate validates a DataPlane object and return the first validation error found.
+func (v *Validator) Validate(dataplane *operatorv1alpha1.DataPlane) error {
+	err := v.ValidateDeployOptions(dataplane.Namespace, &dataplane.Spec.DeploymentOptions)
+	if err != nil {
+		return err
+	}
+	return nil
```
pmalek, line 33:
```suggestion
	return v.ValidateDeployOptions(dataplane.Namespace, &dataplane.Spec.DeploymentOptions)
```

hack/generate-certificates-openssl.sh
```diff
@@ -0,0 +1,36 @@
...
+subjectAltName=@alt_names" > ${cert_dir}/webhook_ext.cnf
+
+openssl genrsa -out ${cert_dir}/ca.key 2048
+openssl req -x509 -new -nodes -key ${cert_dir}/ca.key -out  ${cert_dir}/ca.crt\
+ -subj "/CN=gateway-operator-webhook.kong-system"
+openssl genrsa -out ${cert_dir}/tls.key 2048
+openssl req -new -key  ${cert_dir}/tls.key -out ${cert_dir}/webhook.csr\
+ -subj "/C=US/ST=California/L=San Francisco/O=Kong/OU=Org/CN=gateway-operator-webhook.kong-system"
+openssl x509 -req -in ${cert_dir}/webhook.csr -CA ${cert_dir}/ca.crt -CAkey ${cert_dir}/ca.key \
+ -CAcreateserial -out ${cert_dir}/tls.crt -extensions v3_ext -extfile ${cert_dir}/webhook_ext.cnf
```
pmalek, line 36: EOL

hack/generate-certificates-openssl.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
pmalek, line 1: I'm wondering if this wouldn't be beneficial to write in Go to avoid the possible problems of running bash scripts from Go. This way we'll have way more control over subprocess (function) runtime, return codes etc. 🤔

internal/admission/server.go
```diff
@@ -0,0 +1,179 @@
...
+	if req == nil {
+		response.Allowed = false
+		response.Result = &metav1.Status{
+			Code:    http.StatusBadRequest,
+			Reason:  metav1.StatusReasonBadRequest,
+			Message: "empty request",
+			Status:  metav1.StatusFailure,
+		}
+		return &response, nil
+	}
```
pmalek, line 133: The `nil` check can be the very first thing done in this `func`.

```suggestion
	if req == nil {
		return &admissionv1.AdmissionResponse{
			Allowed: false
			Result: &metav1.Status{
				Code:    http.StatusBadRequest,
				Reason:  metav1.StatusReasonBadRequest,
				Message: "empty request",
				Status:  metav1.StatusFailure,
			}
		}, nil
	}
	
	var (
		response admissionv1.AdmissionResponse
		ok = true
		msg string
		deserializer = codecs.UniversalDeserializer()
	)
```

internal/admission/validator.go
```diff
@@ -0,0 +1,24 @@
...
+func (v *validator) ValidateControlPlane(ctx context.Context, controlPlane operatorv1alpha1.ControlPlane) error {
+	return nil
+}
+
+func (v *validator) ValidateDataPlane(ctx context.Context, dataPlane operatorv1alpha1.DataPlane) error {
+	err := v.dataplaneValidator.Validate(&dataPlane)
+	if err != nil {
+		return err
+	}
+	return nil
```
pmalek, line 23:
```suggestion
	return v.dataplaneValidator.Validate(&dataPlane)
```

test/integration/suite_test.go
```diff
@@ -128,8 +134,37 @@ func TestMain(m *testing.M) {
...
+	// prepare for running webhook if we are going to run webhook tests. includes:
+	// - creating self-signed TLS certificates for webhook server
+	// - creating validaing webhook in test environment
+	runWebhookTests = (os.Getenv("RUN_WEBHOOK_TESTS") == "true")
+	if runWebhookTests {
+		// get IP for genrating certificate and for clients to access.
+		if webhookServerIP == "" {
+			var getIPErr error
+			webhookServerIP, err = getFirstNonLoopbackIP()
+			exitOnErr(getIPErr)
```
pmalek, line 146: This block doesn't look right. `getIPErr` is never assigned to.

test/integration/suite_test.go
```diff
@@ -128,8 +134,37 @@ func TestMain(m *testing.M) {
 	exitOnErr(clusters.KustomizeDeployForCluster(ctx, env.Cluster(), gatewayAPIsCRDs))
 	exitOnErr(waitForCRDs(ctx))
 
+	// prepare for running webhook if we are going to run webhook tests. includes:
+	// - creating self-signed TLS certificates for webhook server
+	// - creating validaing webhook in test environment
+	runWebhookTests = (os.Getenv("RUN_WEBHOOK_TESTS") == "true")
+	if runWebhookTests {
+		// get IP for genrating certificate and for clients to access.
```
pmalek, line 142:
```suggestion
		// get IP for generating certificate and for clients to access.
```

test/integration/suite_test.go
```diff
@@ -128,8 +134,37 @@ func TestMain(m *testing.M) {
 	exitOnErr(clusters.KustomizeDeployForCluster(ctx, env.Cluster(), gatewayAPIsCRDs))
 	exitOnErr(waitForCRDs(ctx))
 
+	// prepare for running webhook if we are going to run webhook tests. includes:
+	// - creating self-signed TLS certificates for webhook server
+	// - creating validaing webhook in test environment
+	runWebhookTests = (os.Getenv("RUN_WEBHOOK_TESTS") == "true")
```
pmalek, line 140: Can we extract this whole block responsible for setup of webhook to a separate func, and give it a parameter of `RUN_WEBHOOK_TESTS` env explaining in the comment how to use it? (the `"true"` value will ....)

pmalek (dismissed):

test/integration/suite_test.go
randmonkey: fixed.

internal/admission/validator.go
randmonkey: updated.

test/integration/suite_test.go
randmonkey: fixed

test/integration/suite_test.go
randmonkey: I tried to move all webhook related code into one function, but it had to be separated into 2 parts:
 - generating webhook certificates, which must be run before operator started
 - waiting for webhook to be available, which must be run after operator started. 
So the code related to webhook has to be seperated into 2 parts, or add some sync mechanism to keep the order.

hack/generate-certificates-openssl.sh
randmonkey: fixed

internal/validation/dataplane/validation.go
randmonkey: This is prepared to add more validations. I added comments here.

hack/generate-certificates-openssl.sh
randmonkey: generating a self-signed certificate set by pure Go has much less practice. Seems @shaneutt  has implemented one: https://medium.com/@shaneutt/create-sign-x509-certificates-in-golang-8ac4ae49f903. But generating certificates by openssl command has much more references. So I decided to use `openssl` command here first.

config/default/manager_webhook_certificates_patch.yaml
randmonkey: fixed

config/admission_webhook/kustomization.yaml
randmonkey: fixed

config/admission_webhook/certificate_config.yaml
randmonkey: fixed.

internal/admission/server.go
randmonkey: updated.

Makefile
randmonkey: removed the duplicate one.

test/integration/suite_test.go
pmalek: I might be not understanding things but I believe 

- your first bullet point is covered by the the code in `if` clause below
- second bullet point is already covered by the `if` statement on line 165

My suggestion in that case is to move the first block ("initialization code") into a separate func: similarly as `waitForWebhook` does.

The reason I'm trying to push this sort of refactoring of this is to prevent bloating of `TestMain` here ( which IMO is already getting pretty hard to reason about ). Ideally this should be refactored into calling a handful of named functions at maximum.

test/integration/suite_test.go
```diff
@@ -231,3 +279,37 @@ func waitForCRDs(ctx context.Context) error {
...
+		return err
+	}
+	webhookCertDir = dir
+
+	fmt.Println("INFO: creating certificates in directory", webhookCertDir)
+	cmd := exec.CommandContext(ctx, "../../hack/generate-certificates-openssl.sh", webhookCertDir, webhookServerIP)
+	return cmd.Run()
+}
+
+func waitForWebhook(ctx context.Context, ip string, port int) error {
```
pmalek, line 297: This in my view, shouldn't be called `waitForWebhook` but `isTCPPortOpen` or similar.

I believe this needs some further explanation in a godoc here.

Additionally: can't we use an HTTP GET here? Wouldn't that translate better into "checking if webhook is ready"? Is that because of certificates and the need to not verify them on such request?

test/e2e/utils_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"strings"
+	"testing"
+
+	"github.com/kong/kubernetes-testing-framework/pkg/clusters"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// namespace creates a namespace for a test case and cleans it up after the test finishes.
+func namespace(t *testing.T) (*corev1.Namespace, func()) {
```
pmalek, line 18: - Perhaps we could rename this to something indicating what's this going to do? Like `createNamespaceForTest()` or similar? `namespace` might also be a quite commonly used identifier in tests conflicting with this functions name.
- I believe I have already suggested this: can't we use `t.Cleanup()` here instead of returning the cleanup function? It would make the house keeping automated and less error prone. WDYT?

test/integration/suite_test.go
```diff
@@ -213,7 +252,16 @@ func startControllerManager() {
...
-	exitOnErr(manager.Run(cfg))
+	if controllerManagerOut != "stdout" {
+		out, err := os.CreateTemp(os.TempDir(), "gateway-operator-controller-logs")
+		exitOnErr(err)
+		cfg.Out = out
+		fmt.Printf("INFO: controller output is being logged to %s\n", out.Name())
+		defer out.Close()
+	}
+	go func() {
+		exitOnErr(manager.Run(cfg))
```
pmalek, line 263: Can't we leave this as it was, i.e. not running this in a goroutine and hence requiring the caller to do `go startControllerManager()` which makes it obvious to the caller that an extra goroutine will be spawned?

Oh and wouldn't this break the stdout code above which via `defer out.Close()` will close the stdout's file and prevent logging from happening?

test/e2e/utils_test.go
```diff
@@ -0,0 +1,72 @@
...
+	buf, err := os.ReadFile("../../config/default/kustomization.yaml")
+	if err != nil {
+		return err
+	}
+
+	// append image contents to replace image
+	fmt.Println("INFO: replacing image in kustomization file")
+	appendImageKustomizationContents := []byte(fmt.Sprintf(gatewayOperatorImageKustomizationContents, imageName, imageTag))
+	newBuf := append(buf, appendImageKustomizationContents...)
+	return os.WriteFile("../../config/default/kustomization.yaml", newBuf, os.ModeAppend)
```
pmalek, line 71: I'm really, really, really hesitant about introducing this implicit files manipulation. 

Beside the fact that this will leave the working directory dirty (right?), the file paths that are used here might quickly become outdated. Ideally these tests would get their own templates to manipulate (as `const`s or in dedicated tests directories).

We can create an issue to address this in the future if we're pressed to merge this PR.

test/integration/suite_test.go
randmonkey: Moved the initialization code for webhook to `prepareWebhook` function.

test/integration/suite_test.go
randmonkey: updated the `waitForWebhook` to use an HTTP GET, and any response is judged as ready.

test/integration/suite_test.go
randmonkey: updated.

test/e2e/utils_test.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// namespace creates a namespace for a test case and cleans it up after the test finishes.
+func createNamespaceForTest(t *testing.T) (*corev1.Namespace, func()) {
+	namespaceForTestCase, err := clusters.GenerateNamespace(ctx, env.Cluster(), t.Name())
+	exitOnErr(err)
+
+	cleanup := func() {
+		assert.NoError(t, clusters.CleanupGeneratedResources(ctx, env.Cluster(), t.Name()))
+	}
```
pmalek, line 24: What about using the `t.Cleanup()`? (Github swallowed my comments ): )

test/e2e/utils_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	fmt.Printf("INFO: restore kustomization file from backup file %s\n", backupKustomizationFile)
+	backUpBuf, err := os.ReadFile(backupKustomizationFile)
+	if err != nil {
+		return err
+	}
+
+	return os.WriteFile(kustomizationFile, backUpBuf, os.ModeAppend)
+
```
pmalek, line 111:
```suggestion
```

test/e2e/utils_test.go
```diff
@@ -0,0 +1,112 @@
...
+const gatewayOperatorImageKustomizationContents = `
+images:
+- name: ghcr.io/kong/gateway-operator:main
+  newName: %v
+  newTag: '%v'
+`
+
+// setOperatorImage appends content for replacing image to kustomization file
+// and
+func setOperatorImage() error {
```
pmalek, line 38: I think this comment is not complete, is it?

test/e2e/utils_test.go
pmalek: Ok got it. What would be the situation where you want to decide the leave the namespace? Debugging? That would require us to modify the code in order to leave the namespace and not trigger the returned cleanup function, right? So basically this would be the same as having `t.Cleanup()` in place and also modifying it in source code (e.g. commenting out?) to prevent the cleanup.

I'm going to pass on the idea of changing this for the sole reason that given that we use some sort of a framework like KTF we can then hook this cleanup func into it and do decision making whether to perform the cleanup there.

test/e2e/utils_test.go
pmalek: Nevermind, found it. Let's disregard my comment here.

test/e2e/utils_test.go
randmonkey: - I agree that we should not affect the working directory when we run tests, so I copied the original content to a backup file and restore it back. Kustomize does not support to use an alternative kustomization file in a directory: [kustomize#4659](https://github.com/kubernetes-sigs/kustomize/issues/4659) so we could only use the backup method to keep the work directory clean.
- I think we would **NOT** use own "templates"(in fact, kustomize is not a template: [kustomize introduction](https://kubernetes.io/blog/2018/05/29/introducing-kustomize-template-free-configuration-customization-for-kubernetes/)  ) for each test case. The e2e test is used to verify that our manifests(generated from kustomization from `config` directory) and code is OK for its usage. So all the test cases are using the same kustomization directory.

test/e2e/utils_test.go
randmonkey: I think I have explained.

test/e2e/utils_test.go
randmonkey: updated. Forgot to add it before.

test/e2e/utils_test.go
randmonkey: fixed.

test/e2e/utils_test.go
```diff
@@ -0,0 +1,112 @@
...
+	file, err := os.CreateTemp("", "kustomization-yaml-backup")
+	if err != nil {
+		return "", err
+	}
+
+	defer file.Close()
+	return file.Name(), nil
+}
+
+func restoreustomizationFile() error {
```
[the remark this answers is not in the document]
randmonkey: fixed.

pmalek:
I've started a discussion on files modification in tests in https://github.com/Kong/team-k8s/discussions/230 which I've touched on in several comments here.

Other than that I feel this is OK: apart from being quite big :) I'll let other 👍 / 👎 this.

shaneutt (approved):
Again, I'm sure due to the size of this we're probably going to end up having to come back to some of the pieces, but at this stage in the project it is OK to take a fix-it-forward mentality, and the core of what we're adding here is good. I advocate that we merge and then fix forward :+1:

shaneutt:
CI was being cranky about the license tests. There was no problem, so I admin merged :+1:

PR #50, opened 2022-06-30
fix: rename /api/ into /apis to make `clientset` generation work without hacks

jrsmroz:
Rename /api/v1alpha1 into /api/operator/v1alpha in order to fix generator hack.

Moves `/api` package to `/apis`. 

The generator does not support unnamed api packages. The generator assumed the `v1alpha1` was  an api package representing a `group name` and assumed the version is not provided.

That resulted in wrong  generation of `/pkg/clientset` package, that was version as `internalversion` and required a sed hack in order to work at all.

Seems the `/api` folder is treated in a special way. I've noticed that operators generaly used `/apis` folder instead.

After the renaming and makefile target modification the generated clientset code does not need a sed hack and the version is correctly set to `v1alpha1`.

**Which issue this PR fixes**

As a side note, if the folder is names `/api`, the generated clientset packages name is `core`.

**PR Readiness Checklist**:


jrsmroz:
Seems I broke something in the process of refactoring. Do not merge until I figure this out.

test/integration/gatewayconfiguration_test.go
```diff
@@ -69,8 +69,8 @@ func TestGatewayConfigurationEssentials(t *testing.T) {
 		},
 		Spec: gatewayv1alpha2.GatewayClassSpec{
 			ParametersRef: &gatewayv1alpha2.ParametersReference{
-				Group:     gatewayv1alpha2.Group(gatewayConfig.GroupVersionKind().Group),
-				Kind:      gatewayv1alpha2.Kind(gatewayConfig.Kind),
+				Group:     gatewayv1alpha2.Group(operatorv1alpha1.SchemeGroupVersion.Group),
+				Kind:      gatewayv1alpha2.Kind("GatewayConfiguration"),
```
jrsmroz, line 73: This is required. During debugging I've found that `gatewayConfig` returned by a `gatewayConfig, err := operatorClient.ApisV1alpha1().GatewayConfigurations(namespace.Name).Create(ctx, gatewayConfig, metav1.CreateOptions{})` has a gkv info stripped on purpose.

Only if the clientset is generated without a version provided (thus using `internalversion`), the gkv is not stripped. I have no idea if we can go past that.

shaneutt (dismissed):
Got a couple of conflicts to sort out, but LGTM

jrsmroz:
@shaneutt , there is other possibility too. Instead of having `/apis/v1alpha1`, we can have `/api/operator/v1alpha1`.
Both works. Depends do we wanna stick to the `/api` folder naming.

shaneutt:
> @shaneutt , there is other possibility too. Instead of having `/apis/v1alpha1`, we can have `/api/operator/v1alpha1`. Both works. Depends do we wanna stick to the `/api` folder naming.

I'm personally fine with `/apis/v1alpha1` this matches what projects like [Gateway API](https://github.com/kubernetes-sigs/gateway-api/tree/main/apis) have and it's always changeable before GA if we really need to :+1:

shaneutt (approved):

PR #51, opened 2022-07-01
fix: controlplane owned deployments scaled down to 0 when dataplane is not set

### LINKED ISSUE #32: Zero scale dormant `ControlPlanes`
### Problem Statement

Technically it's possible to create a `ControlPlane` that isn't connected to any `DataPlane`. While a `ControlPlane` is in this state, we should scale it's `Deployment` down to `0` so that it's not wasting resources and making unnecessary noise.

### Acceptance Criteria

- [ ] A `ControlPlane` created OR updated with no `DataPlanes` results in a `Deployment` with no replicas
- [ ] Any update which adds a `DataPlane` reference then scales up the `Deployment`
- [ ] A `ControlPlane` in this "stagnant" state gets marked with a new, specific status indicating the holding state

jrsmroz:

1. When control plane is created without a dataplane, the control plane deployment  is scaled down to 0 replicas.
2. Control plane status conditions now contains the information that dataplane is not set.
3. Prevents control plane pods from constantly failing when kong Admin API (residing on dataplane owned pod) is not available.
4. When data plane is set in control plane object, the environment variables dependant on data plane name and namespace are reset and propagated to deployment, so the scaled up controlplane's pods will be able to connect to the services like Admin API.
5. Refactors control plane tests, extracting commonly used checks as predicates.

 

**Which issue this PR fixes**

Fixes #32

**PR Readiness Checklist**:

### DIFF
--- a/controllers/controlplane_controller.go
+++ b/controllers/controlplane_controller.go
@@ -72,7 +72,7 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		if errors.IsConflict(err) {
 			debug(
 				log,
-				"conflict found when trying to ensure ControlPlane configuration was up to date, retrying",
+				"conflict found when trying to ensure ControlPlane's DataPlane configuration was up to date, retrying",
 				controlplane,
 			)
 			return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil
--- a/controllers/controlplane_controller.go
+++ b/controllers/controlplane_controller.go
@@ -53,20 +53,6 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil // no need to requeue, status update will requeue
 	}
 
-	debug(log, "validating ControlPlane's DataPlane is set", controlplane)
-	changed, dataplaneOK, err := r.validateDataPlaneIsSet(ctx, controlplane)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
-	if changed {
-		if !dataplaneOK {
-			debug(log, "DataPlane not set, deployment for ControlPlane will remain dormant", controlplane)
-		} else {
-			debug(log, "DataPlane was set, deployment for ControlPlane will be provisioned", controlplane)
-		}
-		return ctrl.Result{}, nil // no need to requeue, status update will requeue
-	}
-
 	debug(log, "validating ControlPlane configuration", controlplane)
 	if len(controlplane.Spec.Env) == 0 && len(controlplane.Spec.EnvFrom) == 0 {
 		debug(log, "no ENV config found for ControlPlane resource, setting defaults", controlplane)
@@ -81,19 +67,24 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil // no need to requeue, the update will trigger.
 	}
 
-	if dataplaneOK {
-		debug(log, "dataplane changed, resetting DataPlane connectivity ENV defaults config", controlplane)
-		changed = setDataPlaneDependantServicesDefaults(&controlplane.Spec.ControlPlaneDeploymentOptions, controlplane.Namespace, nil)
-		if changed {
-			if err := r.Client.Update(ctx, controlplane); err != nil {
-				if errors.IsConflict(err) {
-					debug(log, "conflict found when updating ControlPlane resource, retrying", controlplane)
-					return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil
-				}
-				return ctrl.Result{}, err
-			}
-			return ctrl.Result{}, nil // no need to requeue, the update will trigger.
+	debug(log, "validating ControlPlane's DataPlane configuration", controlplane)
+	if err = r.ensureDataPlaneConfiguration(ctx, controlplane, nil); err != nil {
+		if errors.IsConflict(err) {
+			debug(log, "conflict found when updating ControlPlane resource, retrying", controlplane)
+			return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil
 		}
+		return ctrl.Result{}, err
+	}
+
+	debug(log, "validating ControlPlane's DataPlane status", controlplane)
+	changed, dataplaneIsSet, err := r.ensureDataPlaneStatus(ctx, controlplane, nil)
+	if changed {
+		if dataplaneIsSet {
+			debug(log, "DataPlane was set, deployment for ControlPlane will be provisioned", controlplane)
+		} else {
+			debug(log, "DataPlane not set, deployment for ControlPlane will remain dormant", controlplane)
+		}
+		return ctrl.Result{}, nil // no need to requeue, status update will requeue
 	}
 
 	debug(log, "looking for existing deployments for ControlPlane resource", controlplane)
@@ -102,8 +93,8 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 	if mutated {
-		if !dataplaneOK {
-			debug(log, "DataPlane not set, deployment for ControlPlane will remain scaled to 0 replicas", controlplane)
+		if !dataplaneIsSet {
+			debug(log, "DataPlane not set, deployment for ControlPlane is scaled down to 0 replicas", controlplane)
 			return ctrl.Result{}, nil // no need to requeue until dataplane is set
 		}
 		return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil // TODO: remove after https://github.com/Kong/gateway-operator/issues/26
--- a/controllers/controlplane_controller.go
+++ b/controllers/controlplane_controller.go
@@ -52,6 +52,10 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		debug(log, "ControlPlane resource now marked as scheduled", controlplane)
 		return ctrl.Result{}, nil // no need to requeue, status update will requeue
 	}
+	dataplaneSet, err := r.ensureControlPlaneHasDataplane(ctx, controlplane)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
 
 	debug(log, "validating ControlPlane configuration", controlplane)
 	if len(controlplane.Spec.Env) == 0 && len(controlplane.Spec.EnvFrom) == 0 {
@@ -68,15 +72,21 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	debug(log, "looking for existing deployments for ControlPlane resource", controlplane)
-	created, controlplaneDeployment, err := r.ensureDeploymentForControlPlane(ctx, controlplane)
+	mutated, controlplaneDeployment, err := r.ensureDeploymentForControlPlane(ctx, controlplane)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	if created {
+	if mutated {
 		return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil // TODO: remove after https://github.com/Kong/gateway-operator/issues/26
 	}
 
 	// TODO: updates need to update sub-resources https://github.com/Kong/gateway-operator/issues/27
+
+	debug(log, "checking readiness of ControlPlane deployments", controlplane)
+	if !dataplaneSet {
+		debug(log, "dataplane not set, deployment for ControlPlane will remain dormant", controlplane)
+		return ctrl.Result{}, nil
+	}
 
 	debug(log, "checking readiness of ControlPlane deployments", controlplane)
 	if controlplaneDeployment.Status.Replicas == 0 || controlplaneDeployment.Status.AvailableReplicas < controlplaneDeployment.Status.Replicas {
--- a/controllers/controlplane_controller.go
+++ b/controllers/controlplane_controller.go
@@ -66,14 +66,6 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 		return ctrl.Result{}, nil // no need to requeue, status update will requeue
 	}
-	// if !dataplaneOK {
-	// 	debug(log, "DataPlane not set, deployment for ControlPlane will not be provisioned", controlplane)
-	// 	return ctrl.Result{}, nil // no need to requeue until data plane is set
-	// }
-	// if dataplaneOK && changed {
-	// 	debug(log, "DataPlane was set, deployment for ControlPlane will be provisioned", controlplane)
-	// 	return ctrl.Result{}, nil // no need to requeue, status update will requeue
-	// }
 
 	debug(log, "validating ControlPlane configuration", controlplane)
 	if len(controlplane.Spec.Env) == 0 && len(controlplane.Spec.EnvFrom) == 0 {
@@ -110,6 +102,10 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 	if mutated {
+		if !dataplaneOK {
+			debug(log, "DataPlane not set, deployment for ControlPlane will remain scaled to 0 replicas", controlplane)
+			return ctrl.Result{}, nil // no need to requeue until dataplane is set
+		}
 		return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil // TODO: remove after https://github.com/Kong/gateway-operator/issues/26
 	}
 
--- a/controllers/controlplane_controller.go
+++ b/controllers/controlplane_controller.go
@@ -77,7 +77,10 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	debug(log, "validating ControlPlane's DataPlane status", controlplane)
-	changed, dataplaneIsSet, err := r.ensureDataPlaneStatus(ctx, controlplane, nil)
+	changed, dataplaneIsSet, err := r.ensureDataPlaneStatus(ctx, controlplane)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
 	if changed {
 		if dataplaneIsSet {
 			debug(log, "DataPlane was set, deployment for ControlPlane will be provisioned", controlplane)
--- a/controllers/controlplane_controller.go
+++ b/controllers/controlplane_controller.go
@@ -52,10 +52,6 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		debug(log, "ControlPlane resource now marked as scheduled", controlplane)
 		return ctrl.Result{}, nil // no need to requeue, status update will requeue
 	}
-	dataplaneSet, err := r.ensureControlPlaneHasDataplane(ctx, controlplane)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
 
 	debug(log, "validating ControlPlane configuration", controlplane)
 	if len(controlplane.Spec.Env) == 0 && len(controlplane.Spec.EnvFrom) == 0 {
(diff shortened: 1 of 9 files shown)


pmalek:
One comment before we get to the review stage with this PR: if the state of the PR is still in progress - or in other words it's a draft - please make it so by marking it as draft in Github's UI 🙏

jrsmroz:
> One comment before we get to the review stage with this PR: if the state of the PR is still in progress - or in other words it's a draft - please make it so by marking it as draft in Github's UI 🙏

Hah, sorry. After to many years of work with gitlab I'm used to WIP prefix that does the same, but automatically.

controllers/controlplane_controller.go
```diff
@@ -67,12 +67,39 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil // no need to requeue, the update will trigger.
 	}
 
+	debug(log, "validating ControlPlane's DataPlane configuration", controlplane)
+	if err = r.ensureDataPlaneConfiguration(ctx, controlplane, nil); err != nil {
+		if errors.IsConflict(err) {
+			debug(log, "conflict found when updating ControlPlane resource, retrying", controlplane)
```
shaneutt, line 73: We've got three different verbs used here which could add to some confusion for future readers:

1. validate
2. ensure
3. update

we should try to update the language to be consistent from line to line, perhaps:

```suggestion
	debug(log, "ensuring that the ControlPlane DataPlane configuration is up to date", controlplane)
	if err = r.ensureDataPlaneConfiguration(ctx, controlplane, nil); err != nil {
		if errors.IsConflict(err) {
			debug(log, "conflict found when trying to ensure ControlPlane configuration was up to date, retrying", controlplane)
```

I'm not positive that the language is quite what we want either, but just to illustrate the point.

controllers/controlplane_controller.go
```diff
@@ -67,12 +67,39 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	changed, dataplaneIsSet, err := r.ensureDataPlaneStatus(ctx, controlplane)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if changed {
+		if dataplaneIsSet {
+			debug(log, "DataPlane was set, deployment for ControlPlane will be provisioned", controlplane)
+		} else {
+			debug(log, "DataPlane not set, deployment for ControlPlane will remain dormant", controlplane)
+		}
```
shaneutt, line 89: :+1:

controllers/controlplane_controller.go
```diff
@@ -67,12 +67,39 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	debug(log, "looking for existing deployments for ControlPlane resource", controlplane)
-	created, controlplaneDeployment, err := r.ensureDeploymentForControlPlane(ctx, controlplane)
+	mutated, controlplaneDeployment, err := r.ensureDeploymentForControlPlane(ctx, controlplane)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	if created {
+	if mutated {
+		if !dataplaneIsSet {
+			debug(log, "DataPlane not set, deployment for ControlPlane is scaled down to 0 replicas", controlplane)
```
shaneutt, line 100:
```suggestion
			debug(log, "DataPlane not set, deployment for ControlPlane has been scaled down to 0 replicas", controlplane)
```

controllers/controlplane_controller_conditions.go
```diff
@@ -29,4 +29,8 @@ const (
 	// ControlPlaneConditionReasonPodsReady is a reason which indicates how a ControlPlane
 	// reached fully Provisioned status.
 	ControlPlaneConditionReasonPodsReady = "PodsReady"
+
+	// ControlPlaneConditionsReasonNoDataplane is a reason which indicates that no DataPlane
+	// has been provisioned.
+	ControlPlaneConditionReasonNoDataplane = "NoDataplane"
```
shaneutt, line 35: :+1:

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -62,15 +62,94 @@ func (r *ControlPlaneReconciler) ensureControlPlaneIsMarkedProvisioned(
...
+			case !dataplaneIsSet && condition.Reason != ControlPlaneConditionReasonNoDataplane:
+				condition = metav1.Condition{
+					Type:               string(ControlPlaneConditionTypeProvisioned),
+					Reason:             ControlPlaneConditionReasonNoDataplane,
+					Status:             metav1.ConditionFalse,
+					Message:            "DataPlane is not set",
+					ObservedGeneration: controlplane.Generation,
+					LastTransitionTime: metav1.Now(),
+				}
+				updatedConditions = append(updatedConditions, condition)
```
shaneutt, line 86: So it's kinda a pain, but this can technically break: we have a limit of 8 entries per resource set in the API definition. The logic here has to decide which condition(s) to prune. The `Gateway` implementation currently prunes the oldest conditions when a new one needs to be added. I think in time we'll want to add more robust logic for how this is handled and have a decision tree for which conditions need to be present and which don't.

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -62,15 +62,94 @@ func (r *ControlPlaneReconciler) ensureControlPlaneIsMarkedProvisioned(
...
+			case dataplaneIsSet && condition.Reason == ControlPlaneConditionReasonNoDataplane:
+				condition = metav1.Condition{
+					Type:               string(ControlPlaneConditionTypeProvisioned),
+					Reason:             ControlPlaneConditionReasonPodsNotReady,
+					Status:             metav1.ConditionFalse,
+					Message:            "DataPlane was set, ControlPlane resource is scheduled for provisioning",
+					ObservedGeneration: controlplane.Generation,
+					LastTransitionTime: metav1.Now(),
+				}
+				updatedConditions = append(updatedConditions, condition)
```
shaneutt, line 98: Similar to above, we technically need to account for when we would add a condition beyond the limit. For now pruning the oldest may be sane, but in the future we'll probably need to add something more surgical. I'm OK with adding an issue to tackle this fully later.

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -62,15 +62,94 @@ func (r *ControlPlaneReconciler) ensureControlPlaneIsMarkedProvisioned(
...
+}
+
+// -----------------------------------------------------------------------------
+// ControlPlaneReconciler - Spec Management
+// -----------------------------------------------------------------------------
+
+func (r *ControlPlaneReconciler) ensureDataPlaneConfiguration(
+	ctx context.Context,
+	controlplane *operatorv1alpha1.ControlPlane,
+	envDontOverride map[string]struct{},
```
shaneutt, line 118: So I started adding these maps before, but I realized a while ago this isn't how we should be doing this. Ultimately whether or not an env val should be overridden is a function of whether it's a special value or not and whether it's present. Ultimately though no action needs to be taken here because it would bloat the PR with refactoring that this PR is not responsible for, so this is just a note. The follow up for this is https://github.com/Kong/gateway-operator/issues/65.

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -62,15 +62,94 @@ func (r *ControlPlaneReconciler) ensureControlPlaneIsMarkedProvisioned(
...
 
+// ensureDeploymentForControlPlane ensures that a Deployment is created for the
+// ControlPlane resource. Deployment will remain in dormant state until
+// corresponding dataplane is set.
 func (r *ControlPlaneReconciler) ensureDeploymentForControlPlane(
 	ctx context.Context,
 	controlplane *operatorv1alpha1.ControlPlane,
 ) (bool, *appsv1.Deployment, error) {
-	deployments, err := k8sutils.ListDeploymentsForOwner(ctx, r.Client, consts.GatewayOperatorControlledLabel, consts.ControlPlaneManagedLabelValue, controlplane.Namespace, controlplane.UID)
+	numReplicasWhenNoDataplane := int32(0)
```
shaneutt, line 142: This would appear to be better as a `const` as there's no need to evaluate or store this value more than once per runtime lifespan.

shaneutt (changes requested):
Just a few small comments :+1:

controllers/controlplane_controller_reconciler_utils.go
shaneutt: If you'd like, sure, the behavior for now can be to always override on the controller-owned values and never override on others.

controllers/controlplane_controller_reconciler_utils.go
shaneutt: I'm confused: why can't this be a constant?

controllers/controlplane_controller_reconciler_utils.go
jrsmroz: Shane, as it is now I only seen one condition on dataplane, that is provisioned. We don't keep history for provisioned, it is replaced. I know the logic looks like it's adding new condition, but it is not, it's replacing old condition.
So how come we can reach the limit of 8 ?

controllers/controlplane_controller_reconciler_utils.go
shaneutt: Ok I see, you're trying to make a reference to it further down. While that's probably not a huge deal I would still consider it better overall if you made it a constant as this is more fundamentally accurate (nothing should be changing this, it's always going to be zero).

I would suggest instead we _do make this a `const`_ and instead of taking a reference to it directly, we simply use `k8s.io/utils/pointer`, e.g.:

```golang
deployments[0].Spec.Replicas = pointer.Int32Ptr(numReplicasWhenNoDataplane)
```

jrsmroz: Oh, nice trick, will change.

controllers/controlplane_controller.go
shaneutt: I probably wrote those lines :joy: I do think it's reasonable to challenge everything, but if you don't want to that's fine. Consider this comment resolved at your own discretion as it's really a nitpick anyhow.

jrsmroz: For me, if we can make it better it great. As long as we don't break anything :D

controllers/controlplane_controller_reconciler_utils.go
shaneutt: Right, currently we just replace the only condition so we wont break as of when this PR merges, however it could be a bit of a gotcha for future iterations if they add conditions not realizing the problem so taking care of it now more or less as a breadcrumb for future writers could be helpful, but I would also settle for a follow up issue where we remind ourselves to handle the problem in a forward-thinking manner.

controllers/controlplane_controller_reconciler_utils.go
shaneutt: I think what we're actually looking for is the inverse: a global `alwaysOverride` map.

jrsmroz: Sure, but that's other issue as you said :).

controllers/controlplane_controller_reconciler_utils.go
jrsmroz: Let's create follow up issue.

controllers/controlplane_controller_reconciler_utils.go
shaneutt: Well no, we also want to follow the same logic here, this is pseudocode for the logic we want:

```golang
if isSpecial(var) {
    set(var, newValue)
} else {
   if isAlreadySet(var) {
       return fmt.Errorf("var already provided and not special can't override")
   } else {
       set(var, newValue)
   }
```

controllers/controlplane_controller_utils.go
```diff
@@ -66,6 +119,16 @@ func updateEnv(envVars []corev1.EnvVar, name, val string) []corev1.EnvVar {
 	return newEnvVars
 }
 
+func unsetEnv(envVars []corev1.EnvVar, name string) []corev1.EnvVar {
```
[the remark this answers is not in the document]
jrsmroz: That's a good point, the function removes the var from the slice, but it does not have an effect of unsetting anything yet.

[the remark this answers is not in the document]
jrsmroz: Wen with reject. drop is for removing 1st element.

controllers/controlplane_controller_reconciler_utils.go
jrsmroz: https://github.com/Kong/gateway-operator/issues/66

controllers/controlplane_controller_utils.go
```diff
@@ -42,12 +42,65 @@ func setControlPlaneDefaults(spec *operatorv1alpha1.ControlPlaneDeploymentOption
...
+
+func setControlPlaneEnvOnDataPlaneChange(
+	spec *operatorv1alpha1.ControlPlaneDeploymentOptions,
+	namespace string,
+) bool {
+	var changed bool
+
+	dataplaneIsSet := spec.DataPlane != nil && *spec.DataPlane != ""
+
+	if dataplaneIsSet {
```
mlavacca, line 62: Why not gather these two if blocks in a single one and get rid of the above shared boolean expression?

controllers/controlplane_controller_utils.go
```diff
@@ -42,12 +42,65 @@ func setControlPlaneDefaults(spec *operatorv1alpha1.ControlPlaneDeploymentOption
 
 	if spec.DataPlane != nil && *spec.DataPlane != "" {
 		if _, isOverrideDisabled := dontOverride["CONTROLLER_PUBLISH_SERVICE"]; !isOverrideDisabled {
-			spec.Env = updateEnv(spec.Env, "CONTROLLER_PUBLISH_SERVICE", fmt.Sprintf("%s/svc-%s", namespace, *spec.DataPlane))
+			spec.Env = updateEnv(spec.Env, "CONTROLLER_PUBLISH_SERVICE", controllerPublishService(*spec.DataPlane, namespace))
```
mlavacca, line 45: Since with this PR we are going to set these env vars dynamically with the change of the dataplaneName, I think we could get rid of this part here. What do you think?

controllers/controlplane_controller_utils.go
jrsmroz: Ohh, that is a mistake! it should be `if !dataplaneIsSet`

jrsmroz: Yep. it's a leftover from time when there was envDontOverwrite map.

controllers/controlplane_controller_utils.go
jrsmroz: I was thinking about that. But it saves a single create call during initial reconcile iteration. This will be refactored under https://github.com/Kong/gateway-operator/issues/65

shaneutt (approved):
Once other people's review comments are resolved, LGTM.

controllers/controlplane_controller.go
```diff
@@ -67,12 +67,43 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil // no need to requeue, the update will trigger.
 	}
 
+	debug(log, "validating that the ControlPlane DataPlane configuration is up to date", controlplane)
```
pmalek, line 70:
```suggestion
	debug(log, "validating that the ControlPlane's DataPlane configuration is up to date", controlplane)
```

test/integration/controlplane_test.go
```diff
@@ -50,54 +140,22 @@ func TestControlPlaneEssentials(t *testing.T) {
 	cleaner.Add(dataplane)
 
 	t.Log("verifying deployments managed by the dataplane are ready")
-	require.Eventually(t, func() bool {
-		deployments := mustListDataPlaneDeployments(t, dataplane)
-		return len(deployments) == 1 &&
-			deployments[0].Status.AvailableReplicas >= deployments[0].Status.ReadyReplicas
-	}, time.Minute, time.Second)
+	require.Eventually(t, dataPlaneHasActiveDeployment(t, dataplaneName), time.Minute, time.Second)
```
pmalek, line 143: nit: any particular reason that we wait here for 1 minute and in other places 2 minutes?

I was about to suggest to put that `waitTime` into a variable but then I've noticed the differences between the values. If there's no particular reason why those couldn't be equal I'd try to use a `const` for them

Similarly elsewhere.

pmalek (approved):

test/integration/controlplane_test.go
jrsmroz: Mostly copy pasta and debug leftovers. 1 minute deadline should be enough everywhere. I will double checks the time before merging.

jrsmroz:
Heh, committing a suggestion removed the approvals. Did't expect that.

test/integration/controlplane_test.go
jrsmroz: unified those to 1 minute

test/integration/predicates_test.go
```diff
@@ -31,13 +34,118 @@ func controlPlanePredicate(
...
+		dataplane, err := dataPlaneClient.Get(ctx, dataplaneName.Name, metav1.GetOptions{})
 		require.NoError(t, err)
 		return predicate(dataplane)
 	}
 }
+
+// controlPlaneIsScheduled is a helper function for tests that returns a function
+// that can be used to check if a ControlPlane was scheduled.
+// Should be used in conjunction with require.Eventually or assert.Eventually.
+func controlPlaneIsScheduled(t *testing.T, controlplane types.NamespacedName) func() bool {
```
pmalek, line 51: Nit about those predicates: I can see those being used at some point either by e2e or other suite of tests so I believe we might want to move it (not in this PR) to a separate package and exporting those so that they can be reused in other places.

controllers/controlplane_controller.go
```diff
@@ -67,12 +67,43 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil // no need to requeue, the update will trigger.
 	}
 
+	debug(log, "validating that the ControlPlane's DataPlane configuration is up to date", controlplane)
+	if err = r.ensureDataPlaneConfiguration(ctx, controlplane); err != nil {
+		if errors.IsConflict(err) {
+			debug(
+				log,
+				"conflict found when trying to ensure ControlPlane configuration was up to date, retrying",
```
pmalek, line 75:
```suggestion
				"conflict found when trying to ensure ControlPlane's DataPlane configuration was up to date, retrying",
```

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -62,15 +67,91 @@ func (r *ControlPlaneReconciler) ensureControlPlaneIsMarkedProvisioned(
 	return r.Status().Update(ctx, controlplane)
 }
 
+func (r *ControlPlaneReconciler) ensureDataPlaneStatus(
```
pmalek, line 70: This one either needs a comment or ... an argument? What I mean by that is that a name: `ensureDataPlaneStatus` asks for an argument, i.e. ensure status - what status?

controllers/controlplane_controller_utils.go
```diff
@@ -42,12 +42,63 @@ func setControlPlaneDefaults(spec *operatorv1alpha1.ControlPlaneDeploymentOption
...
+func controllerKongAdminURL(dataplaneName, dataplaneNamespace string) string {
+	return fmt.Sprintf("https://svc-%s.%s.svc:%d",
+		dataplaneName, dataplaneNamespace, dataplaneutils.DefaultKongAdminPort)
+}
+
+func controllerPublishService(dataplaneName, dataplaneNamespace string) string {
+	return fmt.Sprintf("%s/svc-%s", dataplaneNamespace, dataplaneName)
+}
+
+func envValue(env []corev1.EnvVar, name string) string {
```
pmalek, line 95:
```suggestion
func envValueByName(env []corev1.EnvVar, name string) string {
```

Perhaps this could help to make it obvious what this `func` does?

test/integration/controlplane_test.go
```diff
@@ -11,12 +11,94 @@ import (
...
+		},
+	}
+
+	t.Log("deploying controlplane resource without dataplane attached")
+	controlplane, err := controlplaneClient.Create(ctx, controlplane, metav1.CreateOptions{})
+	require.NoError(t, err)
+	cleaner.Add(controlplane)
+
+	t.Log("verifying controlplane state reflects lack of dataplane")
+	require.Eventually(t, controlPlaneDetectedNoDataplane(t, controlplaneName), time.Minute, time.Second)
```
pmalek, line 62: Now that most of these wait times are the same can we use a named `const` for those?

test/integration/predicates_test.go
jrsmroz: Certainly not this in PR, it'd already quite big for what it does.

pmalek: Agreed 👍

test/integration/controlplane_test.go
jrsmroz: Changed both, the deadline and tick to consts.

controllers/controlplane_controller_utils.go
```diff
@@ -42,12 +42,63 @@ func setControlPlaneDefaults(spec *operatorv1alpha1.ControlPlaneDeploymentOption
...
+func controllerKongAdminURL(dataplaneName, dataplaneNamespace string) string {
+	return fmt.Sprintf("https://svc-%s.%s.svc:%d",
+		dataplaneName, dataplaneNamespace, dataplaneutils.DefaultKongAdminPort)
+}
+
+func controllerPublishService(dataplaneName, dataplaneNamespace string) string {
+	return fmt.Sprintf("%s/svc-%s", dataplaneNamespace, dataplaneName)
+}
+
+func envValueByName(env []corev1.EnvVar, name string) string {
```
pmalek, line 95: Can we add a comment on this to explain what it does? When is the empty string returned?

controllers/controlplane_controller_utils.go
```diff
@@ -66,6 +117,16 @@ func updateEnv(envVars []corev1.EnvVar, name, val string) []corev1.EnvVar {
 	return newEnvVars
 }
 
+func rejectEnv(envVars []corev1.EnvVar, name string) []corev1.EnvVar {
```
pmalek, line 120: ditto here

controllers/controlplane_controller_utils.go
jrsmroz: done

PR #53, opened 2022-07-04
chore: bump tests dependencies

jrsmroz:

Updates testing dependencies, mainly kubernetes testing framework to `v0.15.1`. KTF `v0.15.1` uses newer version of metallb manifest, that no longer has `imagePullPolicy` set to `Always`. That should speed up integrations tests startup. 

**Which issue this PR fixes**

**PR Readiness Checklist**:


pmalek (approved):
I believe that instead of bumping this manually we should add a dependabot that would do this for us, similarly as it's done in KIC: https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/dependabot.yml and/or make a PR that actually uses the change and highlights why we need it.

jrsmroz:
> I believe that instead of bumping this manually we should add a dependabot that would do this for us, similarly as it's done in KIC: https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/dependabot.yml and/or make a PR that actually uses the change and highlights why we need it.

Yeah, for sure. It's just an ad hoc solution.

PR #56, opened 2022-07-05
fix: Use namespace (as intended) instead of name when manipulating control plane in env setup

jrsmroz:

- Fixes problem introduced in https://github.com/Kong/gateway-operator/commit/e7ec615494f5b7412e327b3db9f700d35e4200f2#diff-c38af0b84f605ebcf275bd2a3c3ec10b957c3611e9e471660fdc8586008c7276R76
- Use namespace (as intended) instead of name when manipulating control plane in env setup 
- Add tests for control plane essentials
- Add utility tests functions for common checks

**Which issue this PR fixes**

**PR Readiness Checklist**:

### DIFF
--- a/controllers/controlplane_controller.go
+++ b/controllers/controlplane_controller.go
@@ -56,7 +56,7 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	debug(log, "validating ControlPlane configuration", controlplane)
 	if len(controlplane.Spec.Env) == 0 && len(controlplane.Spec.EnvFrom) == 0 {
 		debug(log, "no ENV config found for ControlPlane resource, setting defaults", controlplane)
-		setControlPlaneDefaults(&controlplane.Spec.ControlPlaneDeploymentOptions, controlplane.Name, nil)
+		setControlPlaneDefaults(&controlplane.Spec.ControlPlaneDeploymentOptions, controlplane.Namespace, nil)
 		if err := r.Client.Update(ctx, controlplane); err != nil {
 			if errors.IsConflict(err) {
 				debug(log, "conflict found when updating ControlPlane resource, retrying", controlplane)
--- a/test/integration/predicates_test.go
+++ b/test/integration/predicates_test.go
@@ -1,0 +1,43 @@
+//go:build integration_tests
+// +build integration_tests
+
+package integration
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	operatorv1alpha1 "github.com/kong/gateway-operator/api/v1alpha1"
+)
+
+// controlPlanePredicate is a helper function for tests that returns a function
+// that can be used to check if a ControlPlane has a certain state.
+func controlPlanePredicate(
+	t *testing.T,
+	controlplane *operatorv1alpha1.ControlPlane,
+	predicate func(controlplane *operatorv1alpha1.ControlPlane) bool,
+) func() bool {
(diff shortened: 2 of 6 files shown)


controllers/controlplane_controller.go
```diff
@@ -56,7 +56,7 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	debug(log, "validating ControlPlane configuration", controlplane)
 	if len(controlplane.Spec.Env) == 0 && len(controlplane.Spec.EnvFrom) == 0 {
 		debug(log, "no ENV config found for ControlPlane resource, setting defaults", controlplane)
-		setControlPlaneDefaults(&controlplane.Spec.ControlPlaneDeploymentOptions, controlplane.Name, nil)
+		setControlPlaneDefaults(&controlplane.Spec.ControlPlaneDeploymentOptions, controlplane.Namespace, nil)
```
jrsmroz, line 59: ^ TLDR

test/integration/util_test.go
```diff
@@ -33,3 +38,78 @@ func setup(t *testing.T) (*corev1.Namespace, *clusters.Cleaner) {
 
 	return namespace, cleaner
 }
+
```
pmalek, line 41: Rant: why can't Github figure out how to comment on files :( ....

Real comment:

Can we rethink the placement of those `func`s? `util` is a bag that you can throw anything in. While we're at the beginning of the journey of adding tests and helper functions in here let's invest in better naming/files structure etc.

Perhaps we could add a `predicates_test.go` file for predicates?

pmalek (commented):
W.r.t the PR description:

> Control plane controller uses correct namespace name in env setup

I believe we should try to aim at wording those (PR descriptions) to describe what has been changed or what has been fixed instead of making "general" present tense statements.

My proposal here would be to change it to:

> Use namespace (as intended) instead of name when manipulating control plane in env setup

Or anything similar, mentioning that something's changed/was fixed.

test/integration/util_test.go
jrsmroz: Sure, sounds ok.

controllers/controlplane_controller.go
pmalek: BTW: adding a dedicated type alias for namespace would make making this type of errors harder

test/integration/predicates_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	operatorv1alpha1 "github.com/kong/gateway-operator/api/v1alpha1"
+)
+
```
pmalek, line 14: 👍 For this general idea here.

One food for thought: this is a bit weird to use don't you think? I mean, being used like so:

```go
require.Eventually(t, dataPlanePredicate(t, dataplane, func(dataplane *v1alpha1.DataPlane) bool {
	// ...
	}
)
```

We provide a dataplane, but inside `dataPlanePredicate()` we'll also `Get()` it "again". 

It's somehow unintuitive to me that this happens. Unfortunately I don't have a concrete solution for this but at the very least we should document it in the `dataPlanePredicate`'s comment.

pmalek (commented):
Please address the minor rewording comment in PR's description 🙏

test/integration/predicates_test.go
jrsmroz: We get it again because it might have changed (obviously). May change dataplane into `namespace`, `name` string pair. Nor sure if that helps.

jrsmroz: I've mada a change in the function signature. WDYT?

pmalek (approved):

PR #68, opened 2022-07-08
Possibility to disable the leader election

### LINKED ISSUE #55: Leader election cannot be disabled

### Current Behavior

In a testing/debugging environment, I would like to disable the leader election. This should be possible by not setting the `leader-elect` bool flag. The issue is that such a flag defaults to `true`, hence either setting or not setting the `leader-elect` bool var, the leader election is always set.

### Expected Behavior

One should be able to use or not the leader election. There are two possible solutions to this issue:
* leave the current flag, but default it to false (no leader election by default)
* change the `leader-elect` flag to `no-leader-election`, and default it to false (leader election by default).

mlavacca:
The `--leader-elect` flag is replaced by `--no-leader-election`. The leader election is enabled by default and can be disabled by using such a flag.

**Which issue this PR fixes**
Fixes #55 

**PR Readiness Checklist**:

### DIFF
--- a/internal/manager/run.go
+++ b/internal/manager/run.go
@@ -60,9 +60,9 @@ type Config struct {
 var DefaultConfig = Config{
 	MetricsAddr:     ":8080",
 	ProbeAddr:       ":8081",
-	LeaderElection:  true,
 	WebhookPort:     9443,
 	DevelopmentMode: false,
+	LeaderElection:  true,
 }
 
 func Run(cfg Config) error {
--- a/main.go
+++ b/main.go
@@ -30,18 +30,25 @@ import (
 func main() {
 	var metricsAddr string
 	var probeAddr string
-	var enableLeaderElection bool
+	var disableLeaderElection bool
 	var controllerName string
+
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	flag.BoolVar(&enableLeaderElection, "leader-elect", manager.DefaultConfig.LeaderElection,
-		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&disableLeaderElection, "no-leader-election", false,
+		"Disable leader election for controller manager. Disabling this will not ensure there is only one active controller manager.")
 	flag.StringVar(&controllerName, "controller-name", "", "a controller name to use if other than the default, only needed for multi-tenancy")
 
 	developmentModeEnabled := manager.DefaultConfig.DevelopmentMode
 	if v := os.Getenv("CONTROLLER_DEVELOPMENT_MODE"); v == "true" { // TODO: clean env handling https://github.com/Kong/gateway-operator/issues/19
 		fmt.Println("INFO: development mode has been enabled")
 		developmentModeEnabled = true
+	}
+
+	leaderElection := manager.DefaultConfig.LeaderElection
+	if disableLeaderElection {
+		fmt.Println("INFO: leader election has been disabled")
+		leaderElection = false
 	}
 
 	opts := zap.Options{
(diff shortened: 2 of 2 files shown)


internal/manager/run.go
```diff
@@ -60,7 +60,6 @@ type Config struct {
 var DefaultConfig = Config{
 	MetricsAddr:     ":8080",
 	ProbeAddr:       ":8081",
-	LeaderElection:  true,
```
pmalek, line 63: Might as well make it explicit `false`. The net effect will be the same but a reader not familiar with this struct will not miss the fact that it's set to `false` by default

main.go
```diff
@@ -30,11 +30,11 @@ import (
 func main() {
 	var metricsAddr string
 	var probeAddr string
-	var enableLeaderElection bool
+	var disableLeaderElection bool
 	var controllerName string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	flag.BoolVar(&enableLeaderElection, "leader-elect", manager.DefaultConfig.LeaderElection,
```
pmalek, line 37: BTW: why change the default source?

pmalek:
Quick note: this technically is a breaking change: it changes the end user interface

mlavacca:
> Quick note: this technically is a breaking change: it changes the end user interface

Since we are in pre-alpha I don't see it as a problem at all. Do you?

main.go
mlavacca: because having a bool flag that defaults to true doesn't make any sense

pmalek:
> Since we are **in pre-alpha** I don't see it as a problem at all. Do you?

Right - :smart: - forgot about it. 👍

internal/manager/run.go
mlavacca: My intention is to make explicit the fact that it is not used unless manually set. Even because the leader election is set by default (and can be disabled through the specific flag)

pmalek: Hm 🤔 that's an interesting approach which I like but OTOH `bool` has only 2 possible values, so as I mentioned above - the net effect is the same, regardless if you set it explicitly to false or not set it at all.

I'm not going to push this change but just wanted to make a stance that it will be more obvious to understand (and e.g. find fields of this struct) with this field being set explicitly.

> My intention is to make explicit the fact that **it is not used unless manually set**.

but in line 59 **it is** being used regardless of the value, isn't it?

main.go
mlavacca: and the flag now has the opposite meaning: by setting it you disable the leader election.

internal/manager/run.go
mlavacca: I re-arranged the code to make it clearer, PTAL.

pmalek (approved):

jrsmroz:
Seem like this PR broke the main.

```console
$ k logs gateway-operator-controller-manager-7d45bc54fd-fkf8l -n kong-system
flag provided but not defined: -leader-elect
Usage of /manager:
  -controller-name string
    	a controller name to use if other than the default, only needed for multi-tenancy
  -health-probe-bind-address string
    	The address the probe endpoint binds to. (default ":8081")
  -kubeconfig string
    	Paths to a kubeconfig. Only required if out-of-cluster.
  -metrics-bind-address string
    	The address the metric endpoint binds to. (default ":8080")
  -no-leader-election
    	Disable leader election for controller manager. Disabling this will not ensure there is only one active controller manager.
  -zap-devel
    	Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)
  -zap-encoder value
    	Zap log encoding (one of 'json' or 'console')
  -zap-log-level value
    	Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', or any integer value > 0 which corresponds to custom debug levels of increasing verbosity
  -zap-stacktrace-level value
    	Zap Level at and above which stacktraces are captured (one of 'info', 'error', 'panic').
  -zap-time-encoding value
    	Zap time encoding (one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano'). Defaults to 'epoch'.
```

PR #70, opened 2022-07-08
feat: add data plane network policy restricting admin api to control plane nodes

jrsmroz:
add data plane network policy restricting admin api to control plane nodes

**Which issue this PR fixes**

**PR Readiness Checklist**:


test/integration/predicates_test.go
```diff
@@ -143,6 +148,101 @@ func dataPlaneHasActiveService(t *testing.T, dataplaneName types.NamespacedName,
 	})
 }
 
+func gatewayIsScheduled(t *testing.T, gatewayNSN types.NamespacedName) func() bool {
+	client := gatewayClient.GatewayV1alpha2().Gateways(gatewayNSN.Namespace)
+	return func() bool {
+		gateway, err := client.Get(ctx, gatewayNSN.Name, metav1.GetOptions{})
```
pmalek, line 154: I'd really prefer us not to share objects like `context.Context`. I do see merits in "abusing" it in test suite use case to a degree (oh well, that's how it has to be done in Golang due to lack of better mechanisms?) but in here I'd much rather see us pass the context down from the caller.

Similar thing in other added functions below.

test/integration/predicates_test.go
```diff
@@ -143,6 +148,101 @@ func dataPlaneHasActiveService(t *testing.T, dataplaneName types.NamespacedName,
...
+	return func() bool {
+		networkpolicies, err := gatewayutils.ListNetworkPoliciesForGateway(ctx, mgrClient, gateway)
+		if err != nil {
+			return false
+		}
+		return len(networkpolicies) > 0
+	}
+}
+
+func getShouldReturnResponse(t *testing.T, url string, expectedResponseBody string) func() bool {
```
pmalek, line 217: This doesn't mention anything about "containing" which is actually use in the assertion at the end of returned predicate func. Either we change the name or the assertion (for checking the equality).

test/integration/predicates_test.go
jrsmroz: All in integration tests share the context. It's more to change than just this function. Do not want to change that in this PR.

pmalek: I can see that (: What I suggest here is to change the added functions here and pass down the context from the test itself (possibly handling cancelation of adding timeout to the context if need be).

jrsmroz: Change the function to propagate context

test/integration/predicates_test.go
jrsmroz: True. Changed to `getResponseBodyContains`

controllers/gateway_controller.go
```diff
@@ -213,6 +213,12 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil
 	}
 
+	// ensure the network policy for the dataplane is created
+	debug(log, "ensuring dataplane network policy is in place", gateway)
+	if err := r.ensureDataPlaneNetworkPolicy(ctx, gateway, dataplane, controlplane); err != nil {
+		return ctrl.Result{}, err
+	}
```
shaneutt, line 220: As it stands, this creates (or updates) the `NetworkPolicy` needed for the `ControlPlane` to `DataPlane` traffic, but it doesn't deal with triggering an enqueue of the `Gateway` if the `NetworkPolicy` is deleted or updated elsewhere. https://github.com/Kong/gateway-operator/pull/91 handles this, but as it's open at the same time is this is unaware of this new sub-resource. We should ensure that we coordinate and get this functionality in place.

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -191,3 +194,95 @@ func (r *GatewayReconciler) getGatewayConfigForGatewayClass(ctx context.Context,
...
+	dataplane *operatorv1alpha1.DataPlane,
+	controlplane *operatorv1alpha1.ControlPlane,
+) error {
+	networkPolicies, err := gatewayutils.ListNetworkPoliciesForGateway(ctx, r.Client, gateway)
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return err
+		}
+	}
+
```
shaneutt, line 210: I notice we're not doing anything for the condition where there's more than one `NetworkPolicy` (which, for instance might theoretically be possible if the user turns off leadership election or decides to be "funny" and create and attach a `NetworkPolicy` of their own design). Since the initial POC we've at least thrown an error when this happens, with the understanding that we will have to resolve for these kinds of conditions in time. I suggest we do that at minimum, or (better yet) reduce duplicates.

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -191,3 +194,95 @@ func (r *GatewayReconciler) getGatewayConfigForGatewayClass(ctx context.Context,
...
+				MatchLabels: map[string]string{
+					"kubernetes.io/metadata.name": dataplane.Namespace,
+				},
+			},
+		}},
+	}
+
+	allowProxyIngress := networkingv1.NetworkPolicyIngressRule{
+		Ports: []networkingv1.NetworkPolicyPort{
+			{Protocol: &protocolTCP, Port: &proxyPort},
```
shaneutt, line 256: I think this is OK for now, but at some point we actually should _not_ be allowing raw HTTP traffic anyway. It might be good to add an acceptance criteria for #7 to update this as a part of its scope, and link that issue as a TODO comment here.

internal/consts/consts.go
```diff
@@ -68,6 +68,17 @@ const (
...
 
+// -----------------------------------------------------------------------------
+// Consts - DataPlane exposed ports
+// -----------------------------------------------------------------------------
+
+const (
+	DataPlaneAdminAPIPort = 8444
+	DataPlaneProxyPort    = 8000
+	DataPlaneProxySSLPort = 8443
+	DataPlaneMetricsPort  = 8100
```
shaneutt, line 79: lint: public symbols should have `godoc` comments

internal/utils/gateway/ownerrefs.go
```diff
@@ -124,3 +125,34 @@ func GetDataplaneServiceNameForControlplane(
 
 	return services[0].Name, nil
 }
+
+func ListNetworkPoliciesForGateway(
```
shaneutt, line 129: lint: public symbols should have `godoc` comments.

(we must not have the linter turned on for this :thinking:)

test/integration/gateway_test.go
```diff
@@ -70,80 +70,30 @@ func TestGatewayEssentials(t *testing.T) {
...
-				if condition.Type == string(controllers.DataPlaneConditionTypeProvisioned) && condition.Status == metav1.ConditionTrue {
-					dataplane = &dataplanes[0]
-					return true
-				}
-			}
-		}
-		return false
-	}, subresourceReadinessWait, time.Second)
-	require.NotNil(t, dataplane)
+	require.Eventually(t, gatewayDataPlanesIsProvisioned(t, gateway), subresourceReadinessWait, time.Second)
```
shaneutt, line 84: nit: `gatewayDataPlanesAreProvisioned` OR `gatewayDataPlaneIsProvisioned`

test/integration/gateway_test.go
```diff
@@ -70,80 +70,30 @@ func TestGatewayEssentials(t *testing.T) {
...
-				if condition.Type == string(controllers.ControlPlaneConditionTypeProvisioned) && condition.Status == metav1.ConditionTrue {
-					controlplane = &controlplanes[0]
-					return true
-				}
-			}
-		}
-		return false
-	}, subresourceReadinessWait, time.Second)
-	require.NotNil(t, controlplane)
+	require.Eventually(t, gatewayControlPlanesIsProvisioned(t, gateway), subresourceReadinessWait, time.Second)
```
shaneutt, line 88: nit: `gatewayControlPlanesAreProvisioned` OR `gatewayControlPlaneIsProvisioned`

shaneutt (changes requested):
A few minor requests for change, but looking good :+1: 

One side note: I noticed we added tests that _verify the NetworkPolicy_ is created and deleted, but not that it's specifically being enforced. Thus I assume that this was testing manually against a cluster which enforces `NetworkPolicy`?

Ideally I think we would add to KTF the ability to say:

```go
builder := kind.NewBuilder().WithCalicoCNI()
```

and in our E2E tests call this to ensure that we have a `NetworkPolicy` enforcing CNI, and then have a test that verifies that the `NetworkPolicy` is working as intended (e.g. verify that the `ControlPlane` is able to connect, but errant `Pods` elsewhere can't) so as to avoid us accidentally introducing any regressions to this policy in the future. Let me know your thoughts?

jrsmroz:
> One side note: I noticed we added tests that _verify the NetworkPolicy_ is created and deleted, but not that it's specifically being enforced. Thus I assume that this was testing manually against a cluster which enforces `NetworkPolicy`

Yes. It's mentioned in the PR description.

jrsmroz:
> Ideally I think we would add to KTF the ability to say:
> 
> ```go
> builder := kind.NewBuilder().WithCalicoCNI()
> ```
> 
> and in our E2E tests call this to ensure that we have a `NetworkPolicy` enforcing CNI, and then have a test that verifies that the `NetworkPolicy` is working as intended (e.g. verify that the `ControlPlane` is able to connect, but errant `Pods` elsewhere can't) so as to avoid us accidentally introducing any regressions to this policy in the future. Let me know your thoughts?

I was thinking about adding that to KTF. But in the end I skipped that, as it's low priority task currently. Maybe we open an issue ?

shaneutt:
> > Ideally I think we would add to KTF the ability to say:
> > ```go
> > builder := kind.NewBuilder().WithCalicoCNI()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and in our E2E tests call this to ensure that we have a `NetworkPolicy` enforcing CNI, and then have a test that verifies that the `NetworkPolicy` is working as intended (e.g. verify that the `ControlPlane` is able to connect, but errant `Pods` elsewhere can't) so as to avoid us accidentally introducing any regressions to this policy in the future. Let me know your thoughts?
> 
> I was thinking about adding that to KTF. But in the end I skipped that, as it's low priority task currently. Maybe we open an issue ?

Sure, let's open 2 issues: one in the KTF to add calico CNI as an option for kind clusters, and one in this repo to implement that and add a test.

controllers/gateway_controller.go
jrsmroz: Yeah. I also think that if an error occurs here, we should requeue the reconcile with a backoff.

controllers/gateway_controller_reconciler_utils.go
jrsmroz: If the user attaches its own network policy and marks it as owned by a gateway , should we really delete? It's obvious it's wrong state, yet user created it itself.

If we throw an error, we would stop reconciliation loop. That may force the user to fix the condition. But If it's a result of failed leader election, then we should delete it.

WDYT?

controllers/gateway_controller_reconciler_utils.go
jrsmroz: I'm not sold here. If the user have a valid use case for using http instead of https I would not enforce him.

internal/consts/consts.go
jrsmroz: They do have comment above the const that group the ports. It's fine for the linter as far as I know.

internal/utils/gateway/ownerrefs.go
jrsmroz: Oops.

controllers/gateway_controller.go
shaneutt: > Yeah. I also think that if an error occurs here, we should requeue the reconcile with a backoff.

If you mean the error from `ensureDataPlaneNetworkPolicy()`, returning an error will requeue automatically :+1:

controllers/gateway_controller_reconciler_utils.go
jrsmroz: #7 is for admin api tls, not proxy traffic

controllers/gateway_controller_reconciler_utils.go
jrsmroz: I've added an error a few lines below.

shaneutt:
Calico CNI in KTF is now available in [v0.17.0](https://github.com/Kong/kubernetes-testing-framework/releases/tag/v0.17.0)

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -186,3 +189,100 @@ func (r *GatewayReconciler) getGatewayConfigForGatewayClass(ctx context.Context,
...
+) error {
+	networkPolicies, err := gatewayutils.ListNetworkPoliciesForGateway(ctx, r.Client, gateway.Gateway)
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	numNetworkPolicies := len(networkPolicies)
+	if len(networkPolicies) > 1 {
```
pmalek, line 207:
```suggestion
	if numNetworkPolicies > 1 {
```

Either use the variable that we just created or remove the it - the variable - and use `len(networkPolicies)`

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -186,3 +189,100 @@ func (r *GatewayReconciler) getGatewayConfigForGatewayClass(ctx context.Context,
...
+		return fmt.Errorf("%w, got %d, expected 1", operatorerrors.ErrTooManyDataPlaneNetworkPolicies, numNetworkPolicies)
+	}
+
+	if len(networkPolicies) == 0 {
```
pmalek, line 211:
```suggestion
	if numNetworkPolicies == 0 {
```

ditto here

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -186,3 +189,100 @@ func (r *GatewayReconciler) getGatewayConfigForGatewayClass(ctx context.Context,
 		Name:      gatewayClass.Spec.ParametersRef.Name,
 	}, gatewayConfig)
 }
+
+func (r *GatewayReconciler) ensureDataPlaneNetworkPolicy(
+	ctx context.Context,
+	gateway *gatewayDecorator,
+	dataplane *operatorv1alpha1.DataPlane,
+	controlplane *operatorv1alpha1.ControlPlane,
+) error {
```
pmalek, line 198: This is not immediately obvious what this function does given its name and parameter list. Can we at the very least add a godoc?

test/integration/predicates_test.go
```diff
@@ -143,6 +152,96 @@ func dataPlaneHasActiveService(t *testing.T, dataplaneName types.NamespacedName,
...
+func getResponseBodyContains(t *testing.T, ctx context.Context, url string, responseContains string) func() bool {
+	return func() bool {
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+		require.NoError(t, err)
+
+		resp, err := httpc.Do(req)
+		if err != nil {
+			return false
+		}
```
pmalek, line 234: This somehow touches https://github.com/Kong/kubernetes-ingress-controller/issues/2629.

This snippet fails the test case completely when an HTTP request cannot be created but returns `false` (which will trigger a retry when used with `require.Eventually()`) when issuing that request.

Is that on purpose?

test/integration/asserts_test.go
```diff
@@ -59,3 +77,18 @@ func mustListDataPlaneServices(t *testing.T, dataplane *operatorv1alpha1.DataPla
...
+
+func mustListDataPlanesForGateway(t *testing.T, gateway *v1alpha2.Gateway) []operatorv1alpha1.DataPlane {
+	dataplanes, err := gatewayutils.ListDataPlanesForGateway(ctx, mgrClient, gateway)
+	require.NoError(t, err)
+	return dataplanes
+}
+
+// mustGetGateway is a helper function for tests that
+// conveniently gets a gateway by name.
+func mustGetGateway(t *testing.T, gatewayNSN types.NamespacedName) *v1alpha2.Gateway {
```
pmalek, line 89:
```suggestion
// mustGetGateway is a helper function for tests that conveniently gets a gateway by name.
// It will fail the test if getting the gateway fails.
func mustGetGateway(t *testing.T, gatewayNSN types.NamespacedName) *v1alpha2.Gateway {
```

test/integration/asserts_test.go
```diff
@@ -59,3 +77,18 @@ func mustListDataPlaneServices(t *testing.T, dataplane *operatorv1alpha1.DataPla
...
+	gateways := gatewayClient.GatewayV1alpha2().Gateways(gatewayNSN.Namespace)
+	gateway, err := gateways.Get(ctx, gatewayNSN.Name, metav1.GetOptions{})
```
pmalek, line 91: Shouldn't we pass the `ctx` down from the call site?

test/integration/asserts_test.go
```diff
@@ -59,3 +77,18 @@ func mustListDataPlaneServices(t *testing.T, dataplane *operatorv1alpha1.DataPla
 	require.NoError(t, err)
 	return services
 }
+
+func mustListDataPlanesForGateway(t *testing.T, gateway *v1alpha2.Gateway) []operatorv1alpha1.DataPlane {
+	dataplanes, err := gatewayutils.ListDataPlanesForGateway(ctx, mgrClient, gateway)
```
pmalek, line 82: Ditto w.r.t to the `ctx.`

test/integration/predicates_test.go
jrsmroz: It is on purpose. `NewRequestWithContext` should not fail. it would be a  test author fault, not actual functionality failure. It may be invalid method or nil context.

pmalek: Hmm I'm not sure I have a view on this yet but I just wanted to note that the method cannot be set incorrectly if it's hardcoded in place, like in here using `http.MethodGet` , right?

jrsmroz: If the context is nil, it can be ser incorrectly.

controllers/gateway_controller_reconciler_utils.go
shaneutt: It appears I just got my wires crossed on this one :+1:

shaneutt (changes requested):
Most of this LGTM, but I must be missing something: did we not enable calico for the tests and add a quick verification of the policy's functionality?

jrsmroz:
> Most of this LGTM, but I must be missing something: did we not enable calico for the tests and add a quick verification of the policy's functionality?

We did not. The tests are from pre calico support in KTF.

The test cluster is setup once. That can be done, but let's decide how we want to approach that.
1. The whole integration test suite runs a cluster with calico.
2. The network policy tests run a dedicated cluster with calico.
3. We introduce a testing matrix in the pipeline, rerun all the tests with calico. 
4. Any other ideas?

I opt for option 1 or 2.

pmalek:
> 1. The whole integration test suite runs a cluster with calico.
> 2. The network policy tests run a dedicated cluster with calico.
> 3. We introduce a testing matrix in the pipeline, rerun all the tests with calico.
> 4. Any other ideas?

I'd personally not go for 1 as that would fundamentally change the testing setup that we had until now.

2 would be good but would be nice to have some sort of framework (or concrete approach) for these sorts of "exceptions" (some kind of test matrix).

So perhaps a build tag slapped onto calico related test file(s) and that being used in CI's test matrix?

jrsmroz:
> > 1. The whole integration test suite runs a cluster with calico.
> > 2. The network policy tests run a dedicated cluster with calico.
> > 3. We introduce a testing matrix in the pipeline, rerun all the tests with calico.
> > 4. Any other ideas?
> 
> I'd personally not go for 1 as that would fundamentally change the testing setup that we had until now.
> 
> 2 would be good but would be nice to have some sort of framework (or concrete approach) for these sorts of "exceptions" (some kind of test matrix).
> 
> So perhaps a build tag slapped onto calico related test file(s) and that being used in CI's test matrix?

I don't think it a big deal going with 1 right now. Because all the testing so far was done on kind cluster without custom CNI installed.  We also don't have any release yet.

Calico was sugested because kind cluster does not care about policies at all, thus it's not fully tested when operating on kind without CNI.

pmalek:
> I don't think it a big deal going with 1 right now. Because all the testing so far was done on kind cluster without custom CNI installed. We also don't have any release yet.
> 
> Calico was sugested because kind cluster does not care about policies at all, thus it's not fully tested when operating on kind without CNI.

Agreed but be aware that by doing that we'll basically nuke any support for kind as we won't be testing it in CI. Yes, it's not our primary or even secondary target platform but still. It would be nice to document this with a small note that because kind does not support network policies we're not supporting that functionality there (obvious but worth mentioning IMHO).

jrsmroz:
> > I don't think it a big deal going with 1 right now. Because all the testing so far was done on kind cluster without custom CNI installed. We also don't have any release yet.
> > Calico was sugested because kind cluster does not care about policies at all, thus it's not fully tested when operating on kind without CNI.
> 
> Agreed but be aware that by doing that we'll basically nuke any support for kind as we won't be testing it in CI. Yes, it's not our primary or even secondary target platform but still. It would be nice to document this with a small note that because kind does not support network policies we're not supporting that functionality there (obvious but worth mentioning IMHO).

We do tests on kind, just with custom CNI

pmalek:
> We do tests on kind, just with custom CNI. I'm not sure where to document that. ... Maybe release notes?

I believe a short section with supported (or maybe rather "tested"?) platforms will be fine. And put that info there.

jrsmroz:
> > We do tests on kind, just with custom CNI. I'm not sure where to document that. ... Maybe release notes?
> 
> I believe a short section with supported (or maybe rather "tested"?) platforms will be fine. And put that info there.

As far as I know kind is so far only tested platform. I've added tests for GKE in release pipeline. OpenShift will be somehow tested in future when're done with RedHat certification process.

shaneutt:
I'm actually in favor of option 1. Given that we're only using the most basic features of our CNI currently with kindnet (that is, ip routing) and no tests exist which exercise any other CNI features (honestly, kindnet doesn't have much in the way of extra features) swapping CNI under these circumstances should make no fundamental change. If we were relying on some specific CNI features I might think otherwise. @pmalek you seem to think there's danger here, any specific problems you envision? It's very early in this project's lifespan, could you be convinced that we can just move forward with calico in play for testing?

shaneutt:
I'm not actually opposed to option 1 as we don't rely on any features of our current CNI beyond basic IP routing. That said, I think that a small E2E test where we create an isolated calico enabled cluster would be a fine way to exercise this, the calico tests in KTF could actually probably be copied to some degree.

shaneutt:
I'm not actually opposed to option 1 as we don't rely on any features of our current CNI beyond basic IP routing. That said, I think that a small E2E test where we create an isolated calico enabled cluster would be a fine way to exercise this, the calico tests in KTF could actually probably be copied to some degree. That would seem to indicate option 2. However option 1 is faster and given our current CNI usage I don't think it's going to be any problems, @pmalek could you be convinced or do you feel really strongly against option 1 still? Could you enumerate any of the problems you'd forsee with swapping CNI?

jrsmroz:
@shaneutt Would you mind if we open an issue and do that as a follow up?

pmalek:
> I'm not actually opposed to option 1 as we don't rely on any features of our current CNI beyond basic IP routing. That said, I think that a small E2E test where we create an isolated calico enabled cluster would be a fine way to exercise this, the calico tests in KTF could actually probably be copied to some degree. That would seem to indicate option 2. However option 1 is faster and given our current CNI usage I don't think it's going to be any problems, @pmalek could you be convinced or do you feel really strongly against option 1 still? Could you enumerate any of the problems you'd forsee with swapping CNI?

I'm not going to push against option 1, I just thought that an organized matrix for testing would scale better with time rather than "an exception" of swapped CNIs. But given that we would need to that that anyway to test anything else than basic network features (like e.g. NetworkPolicy) I'm fine with either 1 or 2 and then revisiting if we end up needed another "dimension" for test matrix.

jrsmroz:
> > I'm not actually opposed to option 1 as we don't rely on any features of our current CNI beyond basic IP routing. That said, I think that a small E2E test where we create an isolated calico enabled cluster would be a fine way to exercise this, the calico tests in KTF could actually probably be copied to some degree. That would seem to indicate option 2. However option 1 is faster and given our current CNI usage I don't think it's going to be any problems, @pmalek could you be convinced or do you feel really strongly against option 1 still? Could you enumerate any of the problems you'd forsee with swapping CNI?
> 
> I'm not going to push against option 1, I just thought that an organized matrix for testing would scale better with time rather than "an exception" of swapped CNIs. But given that we would need to that that anyway to test anything else than basic network features (like e.g. NetworkPolicy) I'm fine with either 1 or 2 and then revisiting if we end up needed another "dimension" for test matrix.

I actually opened an issue https://github.com/Kong/gateway-operator/issues/126 that should deal with platforms covarage improvements.

shaneutt (approved):
A follow-up is acceptable :+1:

shaneutt (approved):

jrsmroz:
Follow up - https://github.com/Kong/gateway-operator/issues/133

PR #75, opened 2022-07-12
tests: set up controller logger in integration tests before 30s deadline

### LINKED ISSUE #74: Integrations tests missing controller logs

### Current Behavior

The integrations tests attempt to setup up a logs file for the controller:
```go
func startControllerManager() {
	cfg := manager.DefaultConfig
	cfg.LeaderElection = false
	cfg.DevelopmentMode = true
	cfg.ControllerName = "konghq.com/gateway-operator-integration-tests"

	cfg.NewClientFunc = func(cache cache.Cache, config *rest.Config, options client.Options, uncachedObjects ...client.Object) (client.Client, error) {
		// always hijack and impersonate the system service account here so that the manager
		// is testing the RBAC permissions we provide under config/rbac/. This helps alert us
		// if we break our RBAC configs as the manager will emit permissions errors.
		config.Impersonate.UserName = "system:serviceaccount:kong-system:controller-manager"

		c, err := client.New(config, options)
		if err != nil {
			return nil, err
		}

		return client.NewDelegatingClient(client.NewDelegatingClientInput{
			CacheReader:     cache,
			Client:          c,
			UncachedObjects: uncachedObjects,
		})
	}

	if controllerManagerOut != "stdout" {
		out, err := os.CreateTemp(os.TempDir(), "gateway-operator-controller-logs")
		exitOnErr(err)
		cfg.Out = out
		fmt.Printf("INFO: controller output is being logged to %s\n", out.Name())
		defer out.Close()
	}

	exitOnErr(manager.Run(cfg))
}
```

But it has not effect. Before the `startControllerManager` is even called the default log sink is already set up.

controller runtime has a deadline of `30s` for setting up a logger.
`/sigs.k8s.io/controller-runtime@v0.12.3/pkg/log/log.go`:
```go
// It is safe to assume that if this wasn't set within the first 30 seconds of a binaries
// lifetime, it will never get set. The DelegatingLogSink causes a high number of memory
// allocations when not given an actual Logger, so we set a NullLogSink to avoid that.
//
// We need to keep the DelegatingLogSink because we have various inits() that get a logger from
// here. They will always get executed before any code that imports controller-runtime
// has a chance to run and hence to set an actual logger.
func init() {
	// Init is blocking, so start a new goroutine
	go func() {
		time.Sleep(30 * time.Second)
		loggerWasSetLock.Lock()
		defer loggerWasSetLock.Unlock()
		if !loggerWasSet {
			dlog.Fulfill(NullLogSink{})
		}
	}()
}
```

After `30s` it's impossible to swap the logger. Calls to `SetLogger` have no effect:
```go
// Fulfill switches the logger over to use the actual logger
// provided, instead of the temporary initial one, if this method
// has not been previously called.
func (l *DelegatingLogSink) Fulfill(actual logr.LogSink) {
	if l.promise != nil {
		l.promise.Fulfill(actual)
	}
}
```

Unfortunately the setup of kind  cluster in integration tests takes more than `30s`., Thus the logs are always missing

### Expected Behavior

Integrations tests should catch controller logs to a file.

### Steps To Reproduce

```markdown
1. Run the integrations tests.
2. Look for a line similar to:
`INFO: controller output is being logged to /var/folders/jk/kl7d37ws70j5g_tlwpn41wy80000gn/T/gateway-operator-controller-logs3360770356`
3. Inspect the file content. It will be empty.
```

jrsmroz:

`sigs.k8s.io/controller-runtime` package has a deadline od `30s`, to set up a logger. After the time passes, the logger is set to nil logger and cannot be changed. 

This PR changes the integration tests setup order, so the controller logger is setup before setup of kind cluster begins. Kind cluster setup can take way more than `30s`, thus the logger needs to be in place before the kind cluster is created.

**Which issue this PR fixes**

Fixes #74

Tested manually by running integration tests

**PR Readiness Checklist**:

### DIFF
--- a/test/integration/suite_test.go
+++ b/test/integration/suite_test.go
@@ -164,7 +164,7 @@ func setupControllerLogger() (closeLogFile func() error) {
 	var destWriter io.Writer = os.Stdout
 
 	if controllerManagerOut != "stdout" {
-		out, err := os.CreateTemp(os.TempDir(), "gateway-operator-controller-logs")
+		out, err := os.CreateTemp("", "gateway-operator-controller-logs")
 		exitOnErr(err)
 		fmt.Printf("INFO: controller output is being logged to %s\n", out.Name())
 		destWriter = out
--- a/test/integration/suite_test.go
+++ b/test/integration/suite_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"testing"
@@ -20,8 +21,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 	gatewayclient "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned"
 
@@ -75,6 +78,9 @@ var (
 func TestMain(m *testing.M) {
 	ctx, cancel = context.WithCancel(context.Background())
 	defer cancel()
+
+	closeControllerLogFile := setupControllerLogger()
+	defer closeControllerLogFile()
 
 	var skipClusterCleanup bool
 	var existingCluster clusters.Cluster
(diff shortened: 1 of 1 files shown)


test/integration/suite_test.go
```diff
@@ -149,6 +155,39 @@ func exitOnErr(err error) {
...
+// setupControllerLogger sets up the controller logger.
+// This functions needs to be called before 30sec after the controller packages
+// is loaded, otherwise the logger will not be initialized.
+// Returns the close function, that will close the log file if one was created.
+func setupControllerLogger() (closeLogFile func() error) {
+	var destFile *os.File
+	var destWriter io.Writer = os.Stdout
+
+	if controllerManagerOut != "stdout" {
+		out, err := os.CreateTemp(os.TempDir(), "gateway-operator-controller-logs")
```
pmalek, line 167: The first argument (`dir`) to `os.CreateTemp()` can actually be empty since:

> If dir is the empty string, CreateTemp uses the default directory for temporary files, as returned by TempDir

https://pkg.go.dev/os#CreateTemp

test/integration/suite_test.go
```diff
@@ -149,6 +155,39 @@ func exitOnErr(err error) {
...
+		exitOnErr(err)
```
pmalek, line 168: I'm not going to push it here but I wonder if we shouldn't drop usage of `os.Exit()` as mentioned previously as this inhibits running deferred functions. It's not that big of a deal given that this is called almost at the very beginning of test suite setup but still. Good practices I believe

jrsmroz: I believe we should drop it. But not in this PR. There are more places that would be reworked. Best to open a separate issue for that.

Issue #66, opened 2022-07-08
Improve Conditions handling for API objects


jrsmroz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem Statement

There is  a limit of 8 entries per resource set in the API definition. Currently, when the condition needs to be updated, the code updates all the conditions that matches. In example:
```go
func (r *ControlPlaneReconciler) ensureControlPlaneIsMarkedProvisioned(
	ctx context.Context,
	controlplane *operatorv1alpha1.ControlPlane,
) error {
	updatedConditions := make([]metav1.Condition, 0)
	for _, condition := range controlplane.Status.Conditions {
		if condition.Type == string(ControlPlaneConditionTypeProvisioned) {
			condition.Status = metav1.ConditionTrue
			condition.Reason = ControlPlaneConditionReasonPodsReady
			condition.Message = "pods for all Deployments are ready"
			condition.ObservedGeneration = controlplane.Generation
			condition.LastTransitionTime = metav1.Now()
		}
		updatedConditions = append(updatedConditions, condition)
	}

	controlplane.Status.Conditions = updatedConditions
	return r.Status().Update(ctx, controlplane)
}
```

The Gateway implementation currently prunes the oldest conditions when a new one needs to be added. That's fine, as there is only single occurrence of given condition. But that's not really readable way to update single condition or safe, as if by any change we duplicate a conditions, we might reach the limit of 8.
Then we would need to decide which condition(s) to prune.

### Proposed Solution

- [ ] Ensure functions that set conditions cannot be misused by adding duplicate condition.
- [ ] Work out how to prune conditions, if limit of 8 is reached.

### Additional information

Add more robust logic for how this is handled and have a decision tree for which conditions need to be present and which don't or equivalent solution.

### Acceptance Criteria

- [ ] Ensure functions that set conditions cannot be misused by adding duplicate condition.
- [ ] Work out how to prune conditions, if limit of 8 is reached.

jrsmroz:
@shaneutt Did I messed anything in the description?

shaneutt:
Nope looks good.

jrsmroz:
Pasting suggestions from Ruben Romero from the Slack chanal.

> I've been looking into the gateway-operator code and I would like to make some comments with my personal experience. It is mainly about the Status Condition management.
>
[...]

PR #78, opened 2022-07-13
Conditions as a snapshot

ruromero:
Signed-off-by: ruromero <rromerom@redhat.com>

Following up the discussion on the linked issue and the slack channel. I suggest the Conditions to be used as a snapshot of the current state so that it is easier to understand and monitor the current state of all the managed resources.

This is the current list of Conditions considered in the Gateway Status.

* Scheduled: existing condition describing that the operator will process the gateway
* DataPlaneReady: If the dataplane resource is ready
* ControlPlaneReady: If the controlPlane resource is ready
* GatewayService: whether the Gateway service is properly configured. At the moment if the ip is empty (as it is in my case) it will show a descriptive error instead of an error trace.
* Ready: overall ready condition that aggregates the other conditions

Example: 

```
- lastTransitionTime: "2022-07-13T12:12:58Z"
  message: this gateway has been picked up by the konghq.com/gateway-operator and will be processed
  reason: Scheduled
  status: "True"
  type: Scheduled
- lastTransitionTime: "2022-07-13T12:12:58Z"
  message: ""
  reason: ResourceReady
  status: "True"
  type: DataPlaneReady
- lastTransitionTime: "2022-07-13T12:12:58Z"
  message: ""
  reason: ResourceReady
  status: "True"
  type: ControlPlaneReady
- lastTransitionTime: "2022-07-13T12:12:58Z"
  message: missing loadbalancer.ingress[0].ip
  reason: GatewayServiceError
  status: "False"
  type: GatewayService
- lastTransitionTime: "2022-07-13T12:12:58Z"
  message: ""
  reason: DependenciesNotReady
  status: "False"
  type: Ready
```

**Which issue this PR fixes**

Fix https://github.com/Kong/gateway-operator/issues/66

Let me know if you require a higher level of code coverage at this point

**PR Readiness Checklist**:


jrsmroz:
It seems `ensureDataPlaneStatus` uses the old way to update the status.

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -33,14 +33,15 @@ func (r *ControlPlaneReconciler) ensureControlPlaneIsMarkedScheduled(
...
-		controlplane.Status.Conditions = append(controlplane.Status.Conditions, metav1.Condition{
+		condition := metav1.Condition{
 			Type:               string(ControlPlaneConditionTypeProvisioned),
 			Reason:             ControlPlaneConditionReasonPodsNotReady,
 			Status:             metav1.ConditionFalse,
 			Message:            "ControlPlane resource is scheduled for provisioning",
 			ObservedGeneration: controlplane.Generation,
 			LastTransitionTime: metav1.Now(),
-		})
+		}
```
jrsmroz, line 43: Shouldn't that be replaced by `CreateCondition` call?

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -51,20 +52,16 @@ func (r *ControlPlaneReconciler) ensureControlPlaneIsMarkedProvisioned(
...
-			condition.LastTransitionTime = metav1.Now()
-		}
-		updatedConditions = append(updatedConditions, condition)
+	condition := metav1.Condition{
+		Type:               string(ControlPlaneConditionTypeProvisioned),
+		Reason:             ControlPlaneConditionReasonPodsReady,
+		Status:             metav1.ConditionTrue,
+		Message:            "pods for all Deployments are ready",
+		ObservedGeneration: controlplane.Generation,
+		LastTransitionTime: metav1.Now(),
```
jrsmroz, line 61: `CreateCondition`

controllers/dataplane_controller_reconciler_utils.go
```diff
@@ -29,14 +30,15 @@ func (r *DataPlaneReconciler) ensureDataPlaneIsMarkedScheduled(
...
+		condition := metav1.Condition{
 			Type:               string(DataPlaneConditionTypeProvisioned),
 			Reason:             DataPlaneConditionReasonPodsNotReady,
 			Status:             metav1.ConditionFalse,
 			Message:            "DataPlane resource is scheduled for provisioning",
 			ObservedGeneration: dataplane.Generation,
 			LastTransitionTime: metav1.Now(),
-		})
+		}
+		k8sutils.SetCondition(condition, dataplane)
```
jrsmroz, line 41: `CreateCondition`

controllers/dataplane_controller_reconciler_utils.go
```diff
@@ -47,19 +49,15 @@ func (r *DataPlaneReconciler) ensureDataPlaneIsMarkedProvisioned(
...
-			condition.LastTransitionTime = metav1.Now()
-		}
-		updatedConditions = append(updatedConditions, condition)
+	condition := metav1.Condition{
+		Type:               string(DataPlaneConditionTypeProvisioned),
+		Reason:             DataPlaneConditionReasonPodsReady,
+		Status:             metav1.ConditionTrue,
+		Message:            "pods for all Deployments are ready",
+		ObservedGeneration: dataplane.Generation,
+		LastTransitionTime: metav1.Now(),
```
jrsmroz, line 58: `CreateCondition `

controllers/gateway_controller_conditions.go
```diff
@@ -0,0 +1,64 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+// -----------------------------------------------------------------------------
+// Gateway - Status Condition Types
+// -----------------------------------------------------------------------------
+
+// GatewayConditionType are condition types for Gateway status conditions.
+type GatewayConditionType string
```
jrsmroz, line 13: I guess we need same for data plane and control plane controllers?

controllers/gateway_controller_conditions.go
```diff
@@ -0,0 +1,64 @@
...
+
+const (
+
+	// GatewayServiceErrorReason the Gateway Service is not properly configured
+	GatewayServiceErrorReason GatewayConditionReason = "GatewayServiceError"
+)
+
+type GatewayDecorator struct {
+	*gatewayv1alpha2.Gateway
+}
```
jrsmroz, line 44: I'm ok with usage of decorator, but I would go with less generic name. That generator purpose seems to be implementing `ConditionsAware`.

jrsmroz (commented):
Thanks for submitting that. I like the idea. I've went through and noticed some places that raised my concerns.

controllers/gateway_controller_conditions.go
```diff
@@ -0,0 +1,64 @@
...
+
+// -----------------------------------------------------------------------------
+// Gateway - Status Condition Reasons
+// -----------------------------------------------------------------------------
+
+// GatewayConditionReason are the condition reasons for Gateway status conditions.
+type GatewayConditionReason string
+
+const (
+
```
shaneutt, line 37:
```suggestion
```

controllers/gateway_controller_conditions.go
```diff
@@ -0,0 +1,64 @@
...
+	// GatewayServiceErrorReason the Gateway Service is not properly configured
+	GatewayServiceErrorReason GatewayConditionReason = "GatewayServiceError"
+)
+
+type GatewayDecorator struct {
```
shaneutt, line 42: Does this type need to be public? And in either case it would seem helpful if we could document it with a godoc comment.

internal/utils/kubernetes/status.go
```diff
@@ -0,0 +1,124 @@
+package kubernetes
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+
```
shaneutt, line 8:
```suggestion
```

shaneutt (changes requested):
Looking good, thank you for your contribution :bow: 

A few minor comments.

controllers/controlplane_controller_reconciler_utils.go
jrsmroz: I'm saying it should be `k8sutils.CreateCondition` call instead of creating struct by hand.

controllers/gateway_controller_conditions.go
ruromero: I think this is not the only interface the decorator will be implementing. I like the name of the pattern it is implementing instead of naming it something like `GatewayWithConditions` or similar, but you're the gatekeepers, you have the final word :smile:

controllers/gateway_controller.go
```diff
@@ -94,92 +96,138 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	debug(log, "marking the gateway as ready", gateway)
+	err = r.ensureGatewayMarkedReady(ctx, gateway)
+	if err != nil {
+		k8sutils.SetCondition(k8sutils.CreateCondition(string(GatewayServiceType), metav1.ConditionFalse, string(GatewayServiceErrorReason), err.Error()), gateway)
+	} else {
+		k8sutils.SetCondition(k8sutils.CreateCondition(string(GatewayServiceType), metav1.ConditionTrue, k8sutils.ResourceReadyReason, ""), gateway)
+	}
+	k8sutils.SetReady(gateway)
+	debug(log, "successfully reconciled", gateway)
+	return ctrl.Result{}, r.Client.Status().Update(ctx, gateway.Gateway)
```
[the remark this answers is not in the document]
ruromero: Uhm, this is how it was before IIRC but I'm fine with making it conditional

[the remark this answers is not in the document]
shaneutt: Well at least not in `main`. Currently in main once we log this entry there's no other reconciliation logic performed thereafter:

https://github.com/Kong/gateway-operator/blob/a817f1f1a67a105fb6dc20e8f3e4439298958fbd/controllers/gateway_controller.go#L222

ruromero:
I am adding another convenience method to update only the state when necessary

controllers/gateway_controller_conditions.go
shaneutt: I agree with Jarek that `Decorator` is very generic. I actually don't think `GatewayWithConditions` is really all that bad, but as you suggest that this will be adding more implementations than just conditions so maybe not.

Since this comment was originally made the type has been given godocs to signal its intention, and it's been made private. So at this point it's easy to refactor since it's not part of any public API I think we can keep the name as-is for the moment.

controllers/gateway_controller_conditions.go
ruromero: I made it private

ruromero:
@shaneutt updated and rebased

shaneutt (dismissed):
There are a couple of failing lints and a type issue in the integration tests to clean up but other than that it's looking good. Once @jrsmroz is :heavy_check_mark: I'm :heavy_check_mark: for merging this. I think there's possibly some room for further tweaks but the fundamentals and tests are there and we're in an early development stage so a lot of this will undergo refactor before our first alpha release so we don't want to spend too many cycles off the critical path anyhow.

Thank you so much for your contribution!

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -21,99 +21,65 @@ const numReplicasWhenNoDataplane = 0
...
+			ControlPlaneConditionReasonPodsNotReady,
+			"ControlPlane resource is scheduled for provisioning",
+		)
+
+		k8sutils.SetCondition(condition, controlplane)
+		return true
 	}
 
-	return false, nil
+	return false
```
jrsmroz, line 40: nitpick: So this function is names `isMarkedScheduled`, but it has a sideffect, it modifies the control-plane conditions. That may be misleasing.

ruromero: You're totally right, renaming back

ruromero:
Updated PR. Let me know once it's OK and I'll do the squash

shaneutt (dismissed):

ruromero:
I'm going through the integration tests

shaneutt (approved):
I think the fundamentals here are solid. Since we're still pre-alpha we have the freedom to take a fix-it-forward attitude for any other changes we could have still suggested, so I'm in favor of that approach so as to not increase the size and burden of handling more in this single PR.

This is great, thanks for all the hard work on this @ruromero :vulcan_salute:

controllers/dataplane_controller_reconciler_utils.go
shaneutt: To avoid having to hold up this PR any longer, I've created a follow up for this: https://github.com/Kong/gateway-operator/issues/101

jrsmroz:
Thanks

PR #79, opened 2022-07-13
fix: namespace checked before being dereferenced

### LINKED ISSUE #76: Controller panics when `gatewayClass.Spec.ParametersRef.Namespace` is not set

### Current Behavior

In the Gateway controller, the `GatewayConfig` for a specific `GatewayClass` is retrieved by means of the `gatewayClass.Spec.ParametersRef.Name` and `gatewayClass.Spec.ParametersRef.Namespace` fields. Since the `Namespace` is `*string`, if not set, the reference leads to the operator's panic.
```go
return gatewayConfig, r.Client.Get(ctx, client.ObjectKey{
		Namespace: string(*gatewayClass.Spec.ParametersRef.Namespace),
		Name:      gatewayClass.Spec.ParametersRef.Name,
	}, gatewayConfig)
```

https://github.com/Kong/gateway-operator/blob/main/controllers/gateway_controller_reconciler_utils.go#L155-L158

### Expected Behavior

The operator shouldn't panic. To fix this issue it should be enough to check that the namespace is not nil. If so, replace this reference with an empty string.

mlavacca:
`gatewayClass.Spec.ParametersRef.Namespace` checked before being dereferenced.

**Which issue this PR fixes**
Fixes #76 

**PR Readiness Checklist**:


controllers/gateway_controller_reconciler_utils.go
```diff
@@ -152,8 +152,12 @@ func (r *GatewayReconciler) getGatewayConfigForGatewayClass(ctx context.Context,
 	}
 
 	gatewayConfig := new(operatorv1alpha1.GatewayConfiguration)
+	var namespace string
+	if gatewayClass.Spec.ParametersRef.Namespace != nil {
```
shaneutt, line 156: For thorough-ness we should check if `ParametersRef` is `nil`:

```suggestion
	if gatewayClass.Spec.ParametersRef != nil && gatewayClass.Spec.ParametersRef.Namespace != nil {
```

shaneutt (changes requested):

controllers/gateway_controller_reconciler_utils.go
shaneutt: I see :thinking: 

So functionally that works for our current purposes but it means that the functionality of `getGatewayConfigForGatewayClass` doesn't have this problem solved in encapsulation (that is to say, by itself in a vacuum) so if other callers were to start calling this method it may not be clear to them that this problem exists, and that could create some potential for bugs for any new callers down the road that wouldn't be abundantly clear to the caller unless they carefully studied other calling code.

Generally we shouldn't have to study the calling code for a function or method to understand how to use it properly: we should aim for the function signature and godoc definition to give us as much as possible and for these kinds of problems to be solved internally so I would still advocate that we make this change here for completeness of the method.

mlavacca: I get it, I'll add the check also here.

pmalek (approved):

pmalek (dismissed):

jrsmroz:
@shaneutt  Can we still proceed when it's nil/empty?

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -170,8 +170,12 @@ func (r *GatewayReconciler) getGatewayConfigForGatewayClass(ctx context.Context,
 	}
 
 	gatewayConfig := new(operatorv1alpha1.GatewayConfiguration)
+	var namespace string
+	if gatewayClass.Spec.ParametersRef != nil && gatewayClass.Spec.ParametersRef.Namespace != nil {
+		namespace = string(*gatewayClass.Spec.ParametersRef.Namespace)
+	}
```
shaneutt, line 176: So on second glance, and after thinking about @jrsmroz's comment it seems kinda weird to try the GET if we don't have a namespace, since `GatewayClass` is cluster scoped so there's no _inherent_ namespace and choosing the default namespace of the kubeconfig also feels kinda wrong.

It seems what we should actually be doing is moving the code from the caller that makes a new empty/default config into this method. That will remove the need for the wrapping method and solve this dilemma. Sorry I didn't think of this on my previous review.

shaneutt (changes requested):

jrsmroz:
Can we not use auto merge option? Please 🙏. It's better to do a detailed review instead of pushing on first accept.

shaneutt (changes requested):
So the current code looks good, but it appears we need to add something to the calling code to handle this situation: the currently existing calling code should be able to tolerate the error and simply provide an empty configuration, similar to what it's already doing, since not having a parametersref is not a fatal condition for Gateway reconcilation.

I recommend we create a distinct error in `/pkg/errors`, perhaps something like:

```golang
var ErrObjectMissingReferences = "object has not reference to related objects"
```

Not necessarily 100% sold on the above one, so just consider that an example, but the point being the calling code could do:

```golang
if errors.Is(err, ErrObjectMissingReferences) {
    return NewGatewayConfiguration()
}
```

Let me know your thoughts? :thinking:

jrsmroz:
> > Can we not use auto merge option? Please pray. It's better to do a detailed review by more than one reviewer instead of pushing on first accept.
> > I'm not saying to wait for 2 approvals, but just give people chance to read the code.
> 
> This is probably something we should bring up in retro and discuss making changes to, as the opposite of what you're suggesting has been the standard on the team for a long time.

Sure, I will bring that up. Not sure if auto merges going to scale with bigger team.

shaneutt:
> > > Can we not use auto merge option? Please pray. It's better to do a detailed review by more than one reviewer instead of pushing on first accept.
> > > I'm not saying to wait for 2 approvals, but just give people chance to read the code.
> > 
> > 
> > This is probably something we should bring up in retro and discuss making changes to, as the opposite of what you're suggesting has been the standard on the team for a long time.
> 
> Sure, I will bring that up. Not sure if auto merges going to scale with bigger team.

Yeah, it really might not. We just need to talk about this and find some consensus.

mlavacca:
> the currently existing calling code should be able to tolerate the error and simply provide an empty configuration, similar to what it's already doing, since not having a parametersref is not a fatal condition for Gateway reconciliation.

A new error has been added, and the parametersRef field is checked only inside `getGatewayConfigForGatewayClass`. when not set, an error of type `ObjectMissingParametersRefError` Is returned and caught by the caller, which creates an empty instance of the configuration (as it was already before).

internal/errors/errors.go
```diff
@@ -30,3 +30,27 @@ func IsDataPlaneNotSet(err error) bool {
...
+	object metav1.Object
+
+	message string
+}
+
+func (err *ObjectMissingParametersRefError) Error() string {
+	return err.message
+}
+
+func NewObjectMissingParametersRef(obj metav1.Object) error {
```
shaneutt, line 46: public symbols should have `godoc` comments

internal/errors/errors.go
```diff
@@ -30,3 +30,27 @@ func IsDataPlaneNotSet(err error) bool {
...
+	return &ObjectMissingParametersRefError{
+		object:  obj,
+		message: fmt.Sprintf("object %s/%s has not reference to related objects", obj.GetNamespace(), obj.GetName()),
+	}
+}
+
+func IsObjectMissingParametersRef(err error) bool {
```
shaneutt, line 53: public symbols should have `godoc` comments

internal/errors/errors.go
```diff
@@ -30,3 +30,27 @@ func IsDataPlaneNotSet(err error) bool {
...
+	var onwRefErr *ObjectMissingParametersRefError
+	return errors.As(err, &onwRefErr)
```
shaneutt, line 55: I just noticed that we're using `errors.As` here as was done above previously. `errors.As` makes modifications to the second operand that don't seem to have any bearing here, so even though this PR didn't start this we should take the time to switch to `errors.Is` here and nearby.

shaneutt (changes requested):
I have a couple of small comments and then I think we're ready to go.

On a side note, I noticed a couple things about this area of code:

1. I don't see strong value in having wrapper functions like `IsObjectMissingParametersRef` over just calling `errors.Is(err, ObjectMissingParametersRefError)`. If we continue this trend we'll add 4+ lines for every error we add to minimally change the semantics of a one line check and I'm just not convinced of the value.
2. I noticed we're using `<Name>Error` in these files, but in our public `pkg/errors/` package we use `Err<Name>`. We should be consistent about how we name our errors.

Neither of these problems originates from this PR so I leave it up to you, if you feel inclined to do refactoring then I would ask that we change these things. If you don't and you wouldn't mind putting in a small follow up issue then that's fine too.

internal/errors/errors.go
mlavacca: As far as I know (and I tried as a double check) `errors.As` is the way to handle custom errors that implement the error interface. If we replace it with `errors.Is`, the condition returns always false. `errors.Is` matches only simpler errors created by `errors.New`, and not custom errors, which require `errors.As`.

shaneutt: Well it wouldn't be the first time that I was confused :joy: 

But consider the following program:

```golang
package main

import (
	"errors"
	"fmt"
)

type ErrA struct{}

func (e ErrA) Error() string {
	return "error a"
}

type ErrB struct{}

func (e ErrB) Error() string {
	return "error b"
}

func main() {
	errA := ErrA{}
	errB := ErrB{}
	fmt.Printf("Is a == a? %t\n", errors.Is(errA, ErrA{}))
	fmt.Printf("Is b == b? %t\n", errors.Is(errB, ErrB{}))
	fmt.Printf("Is a == b? %t\n", errors.Is(errA, ErrB{}))
	fmt.Printf("Is b == a? %t\n", errors.Is(errB, ErrA{}))
}
```

It returns this output:

```console
$ go run main.go
Is a == a? true
Is b == b? true
Is a == b? false
Is b == a? false
```

Am I missing something :thinking:

shaneutt: Well it wouldn't be the first time that I was confused :joy: 

But consider the following program:

```golang
package main

import (
	"errors"
	"fmt"
)

type ErrA struct{}

func (e ErrA) Error() string {
	return "error a"
}

type ErrB struct{}

func (e ErrB) Error() string {
	return "error b"
}

func main() {
	errA := ErrA{}
	errB := ErrB{}
	fmt.Printf("Is a == a? %t\n", errors.Is(errA, ErrA{}))
	fmt.Printf("Is b == b? %t\n", errors.Is(errB, ErrB{}))
	fmt.Printf("Is a == b? %t\n", errors.Is(errA, ErrB{}))
	fmt.Printf("Is b == a? %t\n", errors.Is(errB, ErrA{}))
}
```

It returns this output:

```console
$ go run main.go
Is a == a? true
Is b == b? true
Is a == b? false
Is b == a? false
```

Am I missing something :thinking:

shaneutt: Well it wouldn't be the first time that I was confused :joy: 

But consider the following program:

```golang
package main

import (
	"errors"
	"fmt"
)

type ErrA struct{}

func (e ErrA) Error() string {
	return "error a"
}

func NewErrA() error {
	return ErrA{}
}

type ErrB struct{}

func (e ErrB) Error() string {
	return "error b"
}

func NewErrB() error {
	return ErrB{}
}

func main() {
	errA := NewErrA()
	errB := NewErrB()
	fmt.Printf("Is a == a? %t\n", errors.Is(errA, ErrA{}))
	fmt.Printf("Is b == b? %t\n", errors.Is(errB, ErrB{}))
	fmt.Printf("Is a == b? %t\n", errors.Is(errA, ErrB{}))
	fmt.Printf("Is b == a? %t\n", errors.Is(errB, ErrA{}))
}
```

It returns this output:

```console
$ go run main.go
Is a == a? true
Is b == b? true
Is a == b? false
Is b == a? false
```

Am I missing something :thinking:

mlavacca:
> I don't see strong value in having wrapper functions like IsObjectMissingParametersRef over just calling errors.Is(err, ObjectMissingParametersRefError). If we continue this trend we'll add 4+ lines for every error we add to minimally change the semantics of a one line check and I'm just not convinced of the value.

I see the value of those wrappers in having a one-liner that includes these two instructions
```go
var onwRefErr *ObjectMissingParametersRefError
if errors.As(err, &onwRefErr) {
    ...
}
```
for each error check. My opinion is that we can get rid of them, but the code readability will decrease.

> I noticed we're using <Name>Error in these files, but in our public pkg/errors/ package we use Err<Name>. We should be consistent about how we name our errors.

What do you think if we gather all the errors under the same package and we make uniform the naming? When I created this package under internal I didn't notice that there was already one under `pkg`. Would you use `pkg` or `internal`? I personally prefer `pkg` at this point.

shaneutt:
> > I don't see strong value in having wrapper functions like IsObjectMissingParametersRef over just calling errors.Is(err, ObjectMissingParametersRefError). If we continue this trend we'll add 4+ lines for every error we add to minimally change the semantics of a one line check and I'm just not convinced of the value.
> 
> I see the value of those wrappers in having a one-liner that includes these two instructions
> 
> ```go
> var onwRefErr *ObjectMissingParametersRefError
> if errors.As(err, &onwRefErr) {
>     ...
> }
> ```
> 
> for each error check. My opinion is that we can get rid of them, but the code readability will decrease.

Ok, so this hinges on the `errors.Is` vs `errors.As` thing. If we find that I'm wrong about `errors.Is` then we can leave this be. If I'm right though, I think we should remove these because then it's a 1:1 for lines.
 
> > I noticed we're using Error in these files, but in our public pkg/errors/ package we use Err. We should be consistent about how we name our errors.
> 
> What do you think if we gather all the errors under the same package and we make uniform the naming? When I created this package under internal I didn't notice that there was already one under `pkg`. Would you use `pkg` or `internal`? I personally prefer `pkg` at this point.

Sure let's combine them. I would advise moving everything under `internal/errors` for now because it's easy to move things to public later but the opposite is slightly less true and becomes even less true as releases happen. I would also advocate for the `Err<Name>` convention, as this is helpful for autocomplete tooling to hone in on those symbols in a lexicographical manner.

internal/errors/errors.go
mlavacca: Look like the difference is when you have fields in the struct:
```go
package main

import (
	"errors"
	"fmt"
)

type ErrA struct {
	message string
	number  int
}

func (e ErrA) Error() string {
	return fmt.Sprintf("error %s, number %d", e.message, e.number)
}

func NewErrA(message string, number int) error {
	return &ErrA{
		message: message,
		number:  number,
	}
}

type ErrB struct {
	message string
	number  int
}

func (e ErrB) Error() string {
	return "error b"
}

func NewErrB(message string, number int) error {
	return ErrB{
		message: message,
		number:  number,
	}
}

func main() {
	fmt.Println("errors.Is")
	errA := NewErrA("testA", 123)
	errB := NewErrB("testB", 456)
	fmt.Printf("Is a == a? %t\n", errors.Is(errA, ErrA{}))
	fmt.Printf("Is b == b? %t\n", errors.Is(errB, ErrB{}))
	fmt.Printf("Is a == b? %t\n", errors.Is(errA, ErrB{}))
	fmt.Printf("Is b == a? %t\n", errors.Is(errB, &ErrA{}))

	fmt.Println("errors.As")
	var errACheck *ErrA
	var errBCheck *ErrB
	fmt.Printf("Is a == a? %t\n", errors.As(errA, &errACheck))
	fmt.Printf("Is b == b? %t\n", errors.As(errB, &errBCheck))
}
```

It return the output:
```sh
errors.Is
Is a == a? false
Is b == b? false
Is a == b? false
Is b == a? false
errors.As
Is a == a? true
Is b == b? false
```

In that case, to match the error you have to return a pointer to the error, and check it with `errors.As` (as far as I'm getting 😆 )

shaneutt (approved):
Once @jrsmroz's is :+1: I'm :+1:

internal/errors/gatewayclass.go
```diff
@@ -0,0 +1,38 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// -----------------------------------------------------------------------------
+// GatewayClass - Errors
+// -----------------------------------------------------------------------------
+
+// ErrObjectMissingParametersRef is a custom error that must be used when the
+// .spec.ParametersRef field of the given object is nil
+type ErrObjectMissingParametersRef struct {
```
jrsmroz, line 16: That is not proper name. ErrXXX is for vars. For a type it should be `ObjectMissingParametersRerError`

jrsmroz:
> 2\. I noticed we're using `<Name>Error` in these files, but in our public `pkg/errors/` package we use `Err<Name>`. We should be consistent about how we name our errors.

- `<Name>Error` is for types
- `Err<Name>` is for vars.

jrsmroz:
> * I don't see strong value in having wrapper functions like `IsObjectMissingParametersRef` over just calling `errors.Is(err, ObjectMissingParametersRefError)`. If we continue this trend we'll add 4+ lines for every error we add to minimally change the semantics of a one line check and I'm just not convinced of the value.

I fully agree. We willbloat our error package for no good reason. I believe we should define our errors as vars, i.e
```go
var ErrDataPlaneNotSet = errors.New("no dataplane name set on controlplane")
```

Then if one wants to add more context info to that, just use:
```go
return fmt.Errorf("%w, controlplane = %s/%s", operatorerrors.ErrDataPlaneNotSet, obj.GetNamespace(), obj.GetName())
```

internal/errors/controlplane.go
```diff
@@ -0,0 +1,38 @@
...
+// Error returns the error message
+func (err *ErrDataPlaneNotSet) Error() string {
+	return err.message
+}
+
+// NewDataPlaneNotSet returns a new instance of ErrDataPlaneNotSet
+func NewDataPlaneNotSet(obj metav1.Object) error {
+	return &ErrDataPlaneNotSet{
+		object:  obj,
+		message: fmt.Sprintf("no dataplane name set on controlplan %s/%s spec", obj.GetNamespace(), obj.GetName()),
```
jrsmroz, line 30: A type, should be `controlplane`, not `controlplan`

jrsmroz (changes requested):
Please address the review comments regarding error handling.

internal/errors/controlplane.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// -----------------------------------------------------------------------------
+// Controlplane - Errors
+// -----------------------------------------------------------------------------
+
+// ErrDataPlaneNotSet is a custom error that must be used when a specific OwnerReference
+// is expected to be on an object, but it is not found.
+type ErrDataPlaneNotSet struct {
+	object  metav1.Object
```
jrsmroz, line 17: We don't use that addition information. And I doubt we will have a use for it.

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -169,6 +179,12 @@ func (r *GatewayReconciler) getGatewayConfigForGatewayClass(ctx context.Context,
 			}}
 	}
 
+	if gatewayClass.Spec.ParametersRef.Namespace == nil ||
+		*gatewayClass.Spec.ParametersRef.Namespace == "" ||
+		gatewayClass.Spec.ParametersRef.Name == "" {
+		return nil, fmt.Errorf("GatewayClass %s has invalid ParametersRef: both namespace and name must be provided", gatewayClass.Name)
```
jrsmroz, line 185: Frankly, this can be `operatorerrors.ErrObjectInvalidParametersRef`.

jrsmroz (approved):
LGTM. I've added one more comment, but no need to address that right now. `DataPlaneNotSetError` should be changed too, but again, not in this PR.

controllers/gateway_controller_reconciler_utils.go
mlavacca: Yes, it may be, but I don't see the point in having it. In this case, we should requeue with error, as we do.

mlavacca:
> LGTM. I've added one more comment, but no need to address that right now. `DataPlaneNotSetError` should be changed too, but again, not in this PR.

? what's wrong with `ErrDataPlaneNotSet` now?

jrsmroz:
> > LGTM. I've added one more comment, but no need to address that right now. `DataPlaneNotSetError` should be changed too, but again, not in this PR.
> 
> ? what's wrong with `ErrDataPlaneNotSet` now?

My bad. i didn't notice you changed that too.

controllers/gateway_controller_reconciler_utils.go
jrsmroz: Yep, we can skip that now

PR #90, opened 2022-07-18
ci: use separate go.mod for third party tools

pmalek:

Use separate `go.mod` for third party tools to allow separation of dependencies between tools and production&test code as well as making dependabot tools' version bumping possible.


shaneutt (commented):
Generally LGTM once we get that issue with `golangci-lint` figured out and get the linter passing

PR #91, opened 2022-07-18
Subresources replacement

### LINKED ISSUE #26: Replacement of lost sub-resources
## Problem Statement

Currently if you delete a `Deployment` out from underneath a `Gateway` it wont replace it. Similarly, if you edit a `Deployment` such that it becomes detached from its `Gateway` (e.g. by deleting the owner reference) it wont be replaced. The purpose of this task is to ensure that deleted or "stolen" resources get replaced by all relevant controllers.

## Implementation Notes

You'll want to ensure that sub-resource types like `Deployment` and `Service` have `Watch` calls in the relevant controller such that updates to them trigger reconciliation for the parent object (e.g. deleting a `Service` which was owned by the dataplane controller will trigger reconcilation for the `DataPlane` object which owned it, in turn recreating the `Service`). There are examples of this in this repo already, see `controllers/gateway_controller.go` and also in several of the controllers in [KIC](https://github.com/kong/kubernetes-ingress-controller).

## Acceptance Criteria

- [x] if the `Service` for a `DataPlane` is deleted the relevant controller will re-create it
- [x] if the `Service` for a `DataPlane` is edited to remove owner reference, the relevant controller will re-create it
- [x] if the `Deployment` for a `DataPlane` is deleted the relevant controller will re-create it
- [x] if the `Deployment` for a `DataPlane` is edited to remove owner reference, the relevant controller will re-create it
- [x] if the `Deployment` for a `ControlPlane` is deleted the relevant controller will re-create it
- [x] if the `Deployment` for a `ControlPlane` is edited to remove owner reference, the relevant controller will re-create it
- [x] if the `ServiceAccount` for a `ControlPlane` is deleted the relevant controller will re-create it
- [x] if the `ServiceAccount` for a `ControlPlane` is edited to remove owner reference, the relevant controller will re-create it
- [x] if the `ClusterRole` for a `ControlPlane` is deleted the relevant controller will re-create it
- [x] if the `ClusterRole` for a `ControlPlane` is edited to remove owner reference, the relevant controller will re-create it
- [x] if the `ClusterRoleBinding` for a `ControlPlane` is deleted the relevant controller will re-create it
- [x] if the `ClusterRoleBinding` for a `ControlPlane` is edited to remove owner reference, the relevant controller will re-create it
- [x] if the `ControlPlane` for a `Gateway` is deleted the relevant controller will re-create it
- [x] if the `ControlPlane` for a `Gateway` is edited to remove owner reference, the relevant controller will re-create it
- [x] if the `DataPlane` for a `Gateway` is deleted the relevant controller will re-create it
- [x] if the `DataPlane` for a `Gateway` is edited to remove owner reference, the relevant controller will re-create it

### Additional Notes

- [x] TODO comments which linked to this issue are removed once resolved

mlavacca:
Currently if you delete a Deployment out from underneath a Gateway it wont replace it. Similarly, if you edit a Deployment such that it becomes detached from its Gateway (e.g. by deleting the owner reference) it wont be replaced. The purpose of this task is to ensure that deleted or "stolen" resources get replaced by all relevant controllers.

**Which issue this PR fixes**
Fixes #26 

**PR Readiness Checklist**:


pmalek (commented):
I'll do a more detailed review later on (if none of the team members do it faster) but I'd like to point to 3 things:

- Don't we need a CHANGELOG entry for this PR?
- That PR has a pretty empty description: can we add something to it?
- It links to an issue (with lots of sub tasks). Does it solve all of these? There are all already marked as done 🤔

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -166,119 +166,159 @@ func (r *ControlPlaneReconciler) ensureDeploymentForControlPlane(
 
 	count := len(deployments)
 	if count > 1 {
-		return false, nil, fmt.Errorf("found %d deployments for ControlPlane currently unsupported: expected 1 or less", count)
+		if err := r.Client.DeleteAllOf(ctx, &appsv1.Deployment{},
+			client.InNamespace(controlplane.Namespace),
+			client.MatchingLabels{consts.GatewayOperatorControlledLabel: consts.ControlPlaneManagedLabelValue},
```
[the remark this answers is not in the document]
mlavacca: Yep I agree, I'll add some comments

controllers/controlplane_controller_watch.go
```diff
@@ -0,0 +1,123 @@
...
+
+// -----------------------------------------------------------------------------
+// ControlplaneReconciler - Watch Map Funcs
+// -----------------------------------------------------------------------------
+
+func (r *ControlPlaneReconciler) getControlplaneForClusterRole(obj client.Object) (recs []reconcile.Request) {
+	clusterRole, ok := obj.(*rbacv1.ClusterRole)
+	if !ok {
+		log.FromContext(context.Background()).Error(
+			fmt.Errorf("unexpected object type provided"),
```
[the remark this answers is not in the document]
mlavacca: good point, I'll do it

controllers/controlplane_controller_watch.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// -----------------------------------------------------------------------------
+// ControlplaneReconciler - Watch Map Funcs
+// -----------------------------------------------------------------------------
+
+func (r *ControlPlaneReconciler) getControlplaneForClusterRole(obj client.Object) (recs []reconcile.Request) {
+	clusterRole, ok := obj.(*rbacv1.ClusterRole)
+	if !ok {
+		log.FromContext(context.Background()).Error(
```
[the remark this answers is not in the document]
pmalek: Perhaps there's no other way to initialize the logger (as far as I can see in https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.3/pkg/log) 🤔 ?

In such case I'd postpone changing this till a subsequent PR where the `context.Context` is passed down form the call site. The (possibly) network calls like `r.Client.List()` in this file will also benefit from it.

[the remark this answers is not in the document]
shaneutt: > was any specific reason behind having log.FromContext(context.Background()) in the predicates and mapping functions?

The conditions where we would actually log these lines are effectively _programmer error_ type conditions, because they have to do with the wrong type being set. These log lines are there for completeness and ultimately to handle the situation where there was a bug so it could be caught without having to crash the watcher.

> Perhaps there's no other way to initialize the logger

At least at the time of writing I don't think I saw any other way, though you're welcome to dig around. So in theory it's asking us to provide a context but this mapper happens within the [linguistic, not Go] context of the controller-runtime machinery so `context.Background()` is probably reasonably accurate unless you have some additional constraints like a timeout you feel need to be set on it (which I don't think I see any reason to do so).

> In such case I'd postpone changing this till a subsequent PR where the context.Context is passed down form the call site. The (possibly) network calls like r.Client.List() in this file will also benefit from it.

We could advocate for this upstream in `controller-runtime`, arguing that these mapper and predicate functions might need to making blocking calls but I do imagine any traction on such a change would be pretty slow to pick up.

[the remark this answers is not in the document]
pmalek: 👍 I do think that bringing this upstream might be very long and slow journey.

In that case I'd just propose to create an issue to introduce `context.Context` arguments around here so that these can be passed around and used in the logger creation and in k8s API requests.

shaneutt:
Given that we don't have the first release yet, I think it's OK for us to build the changelog entry for `v0.1.0` at the end of all other issues, wherein we just list the high level deliverables provided there. Then after that point we should be treating the changelog normally.

controllers/controlplane_controller_watch.go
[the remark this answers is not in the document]
mlavacca: I took a look at how we do in KIC, and the main difference is that we set and initialize a log in the reconciler struct instance at `SetupWithManager` time in all the controllers. That logger is used anytime we need it (also in the predicates and mapping functions). I think we could follow that approach and have a more structured way to access the logger, but I think that's also out of scope for this PR to refactor our logging structure.

If you agree we can follow the same approach we have in KIC, I'll create an issue to track this.

controllers/dataplane_controller_reconciler_utils.go
```diff
@@ -128,28 +128,41 @@ func (r *DataPlaneReconciler) ensureDeploymentForDataPlane(
...
-		return false, nil, err
+		return nil, err
 	}
 
 	count := len(deployments)
 	if count > 1 {
-		return false, nil, fmt.Errorf("found %d deployments for DataPlane currently unsupported: expected 1 or less", count)
+		// if there is more than one Deployment owned by the same DataPlane,
+		// delete all of them and recreate only one as follows below
+		if err := r.Client.DeleteAllOf(ctx, &appsv1.Deployment{},
```
shaneutt, line 138: Now this replacement over reduction is particularly disruptive, this means if we ever get into a position where we have more than one `DataPlane` running, we stomp all of them, stopping traffic from flowing until the replacement comes online. Similar to my above comments about reducing rather than replacing, we should also do what `ReplicaSets` do with `Pods` and make sure the changing is gradual. For instance, if we decide that we have 1 `DataPlane` that's up and running that needs to be deleted, we don't delete it until we have a replacement `DataPlane` which is in the same state.

For the `DataPlane` version of this, I'm a bit more hesitant to say we can do this in a separate iteration, there's a lot more room for problems here because while the `ControlPlane` operates on eventual consistency terms, the `DataPlane` is realtime.

controllers/dataplane_controller_reconciler_utils.go
```diff
@@ -159,19 +172,33 @@ func (r *DataPlaneReconciler) ensureServiceForDataPlane(
...
-		return false, nil, err
+		return nil, err
 	}
 
 	count := len(services)
 	if count > 1 {
-		return false, nil, fmt.Errorf("found %d services for DataPlane currently unsupported: expected 1 or less", count)
+		// if there is more than one Service owned by the same DataPlane,
+		// delete all of them and recreate only one as follows below
+		if err := r.Client.DeleteAllOf(ctx, &corev1.Service{},
```
shaneutt, line 182: Possibly even more disruptive than the `DataPlane` version: this can cause an established IP address for the `DataPlane` to be cycled. We have to try and protect the IP.

shaneutt (changes requested):
My main takeaways here are concerns about our replacement over reduction strategy for dealing with extra sub-resources. In some cases (some of which are edge, some maybe less so) this could cause significant disruption and even outage I think so let's take a good long look at these and see if we can remove those potential disruptions.

controllers/controlplane_controller.go
```diff
@@ -27,9 +35,31 @@ type ControlPlaneReconciler struct {
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *ControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	clusterRolePredicate := predicate.NewPredicateFuncs(r.clusterRoleHasControlplaneOwner)
+	clusterRolePredicate.UpdateFunc = func(e event.UpdateEvent) bool {
+		return r.clusterRoleHasControlplaneOwner(e.ObjectOld)
```
jrsmroz, line 40: Shouldn't this filter only the events that change the cluster role to state that we don't expect instead of all states?

controllers/controlplane_controller_watch.go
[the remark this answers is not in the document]
jrsmroz:
```go
// FromContext returns a logger with predefined values from a context.Context.
func FromContext(ctx context.Context, keysAndValues ...interface{}) logr.Logger {
	log := Log
	if ctx != nil {
		if logger, err := logr.FromContext(ctx); err == nil {
			log = logger
		}
	}
	return log.WithValues(keysAndValues...)
}
```
Will call
```go

func FromContext(ctx context.Context) (Logger, error) {
	if v, ok := ctx.Value(contextKey{}).(Logger); ok {
		return v, nil
	}

	return Logger{}, notFoundError{}
}
```

In the end, the global Log will be used anyway. 
I do not see any benefit over jus using this background thing over

```go
log.Log.Error( //...

```

controllers/gateway_controller.go
```diff
@@ -114,34 +130,12 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 	r.setDataplaneGatewayConfigDefaults(gatewayConfig)
 
-	debug(log, "looking for associated dataplanes", gateway)
-	dataplanes, err := gatewayutils.ListDataPlanesForGateway(
-		ctx,
-		r.Client,
-		gateway,
-	)
+	debug(log, "ensuring DataPlane for Gateway exists", gateway)
+	dataplane, err := r.ensureDataPlaneForGateway(ctx, gateway, gatewayConfig)
```
jrsmroz, line 134: You may want to pull the latest changes from main. Reconcile loop was totally redone when new conditions handling was added.

controllers/controlplane_controller.go
```diff
@@ -27,9 +35,31 @@ type ControlPlaneReconciler struct {
...
+
+	clusterRoleBindingPredicate := predicate.NewPredicateFuncs(r.clusterRoleBindingHasControlplaneOwner)
+	clusterRoleBindingPredicate.UpdateFunc = func(e event.UpdateEvent) bool {
+		return r.clusterRoleBindingHasControlplaneOwner(e.ObjectOld)
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&operatorv1alpha1.ControlPlane{}).
 		Named("ControlPlane").
+		Owns(&appsv1.Deployment{}).
```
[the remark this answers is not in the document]
rainest: Can you elaborate on what we'd have less control over and what the awkward semantics are?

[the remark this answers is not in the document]
shaneutt: I think the main thing is that I like having direct access to the `Watch()` method present in the `controller.Controller` type, but this ultimately I don't think actually limits our control for our purposes. In retrospect and because this PR has been sitting for a long time, I'm going to consider this one resolved because I don't think we need to spend any further time on it as it's more of a nit, and because it's easy to change later if we run into any problems.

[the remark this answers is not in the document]
rainest: I'd still be interested in hearing what you've encountered in the past--you've got more controller-runtime experience than I do and this is the first time I've encountered `Owns()`, so it'd be useful to understand scenarios where you'd be inclined to use it but would need to add `Watch()` parameters beyond whatever the `Owns()` sets up on your behalf.

mlavacca:
Reverting this PR back to draft as it needs to be reworked in regards to the reduction over replacement strategy mentioned by Shane

controllers/controlplane_controller.go
```diff
@@ -108,31 +138,22 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
 	debug(log, "ensuring that ClusterRoleBindings for ControlPlane Deployment exist", controlplane)
-	created, _, err = r.ensureClusterRoleBindingForControlPlane(ctx, controlplane, controlplaneServiceAccount.Name, controlplaneClusterRole.Name)
+	_, err = r.ensureClusterRoleBindingForControlPlane(ctx, controlplane, controlplaneServiceAccount.Name, controlplaneClusterRole.Name)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	if created {
-		return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil // TODO: remove after https://github.com/Kong/gateway-operator/issues/26
-	}
```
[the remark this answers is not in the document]
mlavacca: I'm not sure I'm getting your point here, do you mean we should leave these requeues as they are?

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -166,119 +166,167 @@ func (r *ControlPlaneReconciler) ensureDeploymentForControlPlane(
 
 	count := len(deployments)
 	if count > 1 {
-		return false, nil, fmt.Errorf("found %d deployments for ControlPlane currently unsupported: expected 1 or less", count)
+		// if there is more than one Deployment owned by the same ControlPlane,
+		// delete all of them and recreate only one as follows below
+		if err := r.Client.DeleteAllOf(ctx, &appsv1.Deployment{},
```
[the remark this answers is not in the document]
mlavacca: Gotcha, I didn't consider this aspect. I'll dig in the `ReplicaSet` and `Pod` controllers to find if we can take inspiration for this pattern. If those changes will make this PR grow too much, I'll create follow-up issues.

controllers/dataplane_controller_reconciler_utils.go
mlavacca: Ok, I'll address this reduction problem first, as it has more priority than others.

controllers/gateway_controller.go
mlavacca: yeah, I'm handling tons of conflict, not funny 😞

mlavacca:
@shaneutt since the reduction over replacement strategy would make this PR kinda huge, I think it'd be better to reduce this PR's scope and not handling the situation in which more that one object exists. If you agree, I'll leave all the conditionals `if count > 1 {//do stuff}` and create a follow-up PR to resolve the reduction problem specifically. Does it work for you?

controllers/controlplane_controller.go
mlavacca: Since the ClusterRole is owned by the controlplane, we should watch for all its events (as vanilla Owns does). We could perform more granular filtering, but it is only a matter of optimization and out of this PR's scope. I wrote the reason behind this predicate in a comment, copying here:

```
// we need to get the old objects for cluster-wide resource updates to address
// cases in which the ownerReference is changed or deleted. If we use the newObject,
// we cannot infer the object owner, as there is no reference to it, and we'd lose
// the event that should trigger the creation of a new resource
```

mlavacca:
Here are a couple of follow-up issues:
* #182 
* #183

controllers/controlplane_controller.go
```diff
@@ -32,9 +40,44 @@ type ControlPlaneReconciler struct {
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *ControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	// we need to get the old objects for cluster-wide resource updates to address
+	// cases in which the ownerReference is changed or deleted. If we use the newObject,
+	// we cannot infer the object owner, as there is no reference to it, and we'd lose
+	// the event that should trigger the creation of a new resource
```
shaneutt, line 47: :+1: I appreciate the comment here. I do wonder though if I may suggest a rewording that I think might be a bit more straightforward?

```suggestion
	// for owned objects we need to check if updates to the objects resulted in the
	// removal of an OwnerReference to the parent object, and if so we need to
	// enqueue the parent object so that reconcilation can create a replacement.
```

Just a thought, let me know your thoughts? Not a blocker, resolve at your own discretion.

controllers/controlplane_controller.go
```diff
@@ -32,9 +40,44 @@ type ControlPlaneReconciler struct {
...
+			handler.EnqueueRequestsFromMapFunc(r.getControlplaneForClusterRole),
+			builder.WithPredicates(clusterRolePredicate)).
+		// watch for changes in ClusterRoleBindings created by the controlplane controller.
+		// Since the ClusterRoleBindings are cluster-wide but controlplanes are namespaced,
+		// we need to manually detect the owner by means of the UID
+		// (Owns cannot be used in this case)
+		Watches(
+			&source.Kind{Type: &rbacv1.ClusterRoleBinding{}},
+			handler.EnqueueRequestsFromMapFunc(r.getControlplaneForClusterRoleBinding),
+			builder.WithPredicates(clusterRoleBindingPredicate)).
```
shaneutt, line 80: Good comments! Everything we're doing here is nice and clear at a glance for future readers :+1:

controllers/controlplane_controller.go
```diff
@@ -127,12 +170,12 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
 	debug(log, "creating MTLS certificate", controlplane)
-	created, certSecretName, err := r.ensureCertificate(ctx, controlplane)
+	created, certSecret, err := r.ensureCertificate(ctx, controlplane)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 	if created {
-		return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil // TODO: remove after https://github.com/Kong/gateway-operator/issues/26
+		return ctrl.Result{}, nil
```
shaneutt, line 178: might be worth a breadcrumb comment here for future readers:

```suggestion
		return ctrl.Result{}, nil // requeue will be triggered by the creation of the owned object
```

There's actually several places where something similar happens, I know it might seem excessive but I would personally just put a comment like this on each of them.

controllers/controlplane_controller_watch.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (r *ControlPlaneReconciler) objHasControlplaneOwner(ctx context.Context, obj client.Object) bool {
+	controlplaneList := &operatorv1alpha1.ControlPlaneList{}
+	if err := r.Client.List(ctx, controlplaneList); err != nil {
+		// filtering here is just an optimization, the reconciler will check the
+		// class as well. If we fail here it's most likely because of some failure
+		// of the Kubernetes API and it's technically better to enqueue the object
+		// than to drop it for eventual consistency during cluster outages.
+		log.FromContext(ctx).Error(err, "could not list controlplanes in map func")
+		return true
```
shaneutt, line 62: :+1: we need to keep this in mind going forward, it's kinda subtle: unless you get a type error, you pretty much just wanna return true for these kinds of filters _unless_ you are _certain_ the object is not relevant. This can help to reduce "stuck" behavior of the operator when a cluster API outage occurs by ensuring that the resource gets into the queue and will _eventually_ be reconciled when the API comes back to health.

controllers/gateway_controller.go
```diff
@@ -39,46 +39,30 @@ type GatewayReconciler struct {
...
+			&source.Kind{Type: &operatorv1alpha1.GatewayConfiguration{}},
+			handler.EnqueueRequestsFromMapFunc(r.listGatewaysForGatewayConfig),
+			builder.WithPredicates(predicate.NewPredicateFuncs(r.gatewayConfigurationMatchesController))).
+		// watch for updates to GatewayClasses, if any GatewayClasses change, enqueue
+		// reconciliation for all supported gateway objects which reference it.
+		Watches(
+			&source.Kind{Type: &gatewayv1alpha2.GatewayClass{}},
+			handler.EnqueueRequestsFromMapFunc(r.listGatewaysForGatewayClass),
+			builder.WithPredicates(predicate.NewPredicateFuncs(r.gatewayClassMatchesController))).
+		Complete(r)
```
shaneutt, line 65: I see you really like the other way of initializing the controller. Which is fine, but at this point let's just make sure we're consistent about this across the repo.

controllers/utils.go
```diff
@@ -269,3 +293,36 @@ func deploymentOptionsDeepEqual(opts1, opts2 *operatorv1alpha1.DeploymentOptions
...
+}
+
+func labelObjForOwner(obj client.Object, owner client.Object) {
+	switch owner.(type) {
+	case *operatorv1alpha1.ControlPlane:
+		labelObjForControlPlane(obj)
+	case *operatorv1alpha1.DataPlane:
+		labelObjForDataplane(obj)
+	}
+}
```
shaneutt, line 318: nit: this is a mutator, where the functions around it are retrievers. A really tiny thing, but I would suggest not grouping them together. Perhaps we can just place it right next to the other functions it calls with similar purpose?

Also (and I know this is most likely MY fault so I get the irony) it seems like we should actually put a verb on this function and the underlying functions: we put `get<thing` on functions that retrieve data from some given information, perhaps we should put `addLabelFor<thing>` for these three functions so it's more clear at a glance that it causes object mutation?

Really minor things, but could slightly improve readability.

internal/manager/run.go
```diff
@@ -131,6 +132,18 @@ func Run(cfg Config) error {
...
+		// for the controlplane (or dataplane) and returns with success. In this scenario there
+		// is already another request in the queue that triggers a new reconciliation loop;
+		// when the controller checks whether there are already deployments, no deployment is
+		// found because the cache has not been updated yet. In the following reconicilation
+		// loop, an error occurs because we have two deployments.
+		// TODO: https://github.com/Kong/gateway-operator/issues/182
+		ClientDisableCacheFor: []client.Object{
+			&appsv1.Deployment{},
+			&corev1.Secret{},
+		},
```
shaneutt, line 146: This one is a bit tricky. At certain scales this will definitely cause undo stress on the API server, we've seen it in our other products with `Secrets` some people just have absolutely massive amounts of them. I appreciate your comments on this explaining the situation, albeit I might not be 100% clear on the issue:

We state here that the problem occurs when we create a `Deployment` (for instance), there's already another request for the parent object in the queue? Ideally what should happen is this:

1. we create the `Deployment`
2. we end reconcilation, _without_ requeuing
3. the creation of the `Deployment` results in the parent object being enqueued
4. reconcilation is re-entered and can reach the next step where it can find the `Deployment`

Are there problems with making this work? Am I missing something?

shaneutt (changes requested):
Things are mostly looking good, but I have two explicit blockers:

* we should not be disabling cache for `Deployments` and `Secrets`, we need to try to find an alternative solution for this problem. I might be agreeable to doing a follow-up if it gets to big, but ultimately we shouldn't have this post alpha and if we can solve it now we should
* I don't see tests. Ideally we should add unit tests throughout. Also while I get the idea is that the integration tests should continue to pass, we should actually be adding integration tests where we intentionally delete and "de-own" objects and verify the replacement behavior is working correctly (and not creating too many, OR too little (none) replacements).

controllers/gateway_controller.go
mlavacca: Yep, I changed the Gateway controller setup for this reason :) we should be consistent across the repo at this point.

internal/manager/run.go
mlavacca: I wasn't totally clear in the comment, let me add some more context. 
The situation you depicted as a bullet list reflects 100% what happens. It works 99% percent of the time (I never succeeded in triggering this error on my local machine), but under certain conditions (in CI at least 1 test fails every time, there is no specific failing test, and all of them fail randomly with the same reason) something strange  in the dataplane controller loop happens:
1. the mTLS secret is created by the dataplane controller, and the loop ends
2. the secret creation triggers another loop in the dataplane controller
3. the dataplane controller loop tries to retrieve the mTLS secret by label and `ownerRef`, but it fails. No mTLS secrets belonging to the dataplane are found in the cache.
4. as a result, another mTLS secret is created
5. Another reconciliation loop (triggered by the second mTLS secret creation) happens
6. the dataplane controller lists all the secrets owned by the dataplane, and two secrets are found. 
7. There is no recovery from this situation at the moment.

To have evidence of the issue, try searching for "found 2" in the integration test pipeline in this PR, and you'll see what I'm talking about. In CI, This happens very often with the dataplane secrets and sometimes with the dataplane deployment. If we disable the cache for Deployments and Secret, the tests succeed at 100%. 

I get that disabling the cache for those objects is a workaround to the problem, but I tried to figure out the root cause of the problem for a while without success. When we'll be able to recover from the situation in which two objects of the same type exist, we can re-enable the cache.

Furthermore, I totally understand that in a production environment with plenty of secrets, reading directly from the API server could create a heavy load on it, but since we are still working on a technical preview, do you still think it is unacceptable having this workaround for the moment?

mlavacca:
> I don't see tests. Ideally we should add unit tests throughout. Also while I get the idea is that the integration tests should continue to pass, we should actually be adding integration tests where we intentionally delete and "de-own" objects and verify the replacement behavior is working correctly (and not creating too many, OR too little (none) replacements).

I'm gonna add some tests

internal/manager/run.go
shaneutt: Talked about this one with @mlavacca in Slack: we're going to go with the cache disable option for the moment since we're still in an early alpha phase and it's OK to use this as a workaround for the time being. https://github.com/Kong/gateway-operator/issues/182 is the follow up to solve this wholesale before the next alpha release.

shaneutt (approved):
The new tests look adequate for now so that we avoid regressions :+1:

This seems like a good place to stop and merge, and as discussed we will continue the implementation in the follow-up issues we created here.

PR #92, opened 2022-07-18
docs: add plugins kep

shaneutt:

Adds a KEP to express the motivation and goals to add a plugin interface to the KGO.

This intentionally only includes summary, motivation and goals so that we can build consensus around the "what" and "why" before we concern ourselves with the "how".


keps/0002-plugins.md
```diff
@@ -0,0 +1,49 @@
...
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Design](#design)
+    - [API Resources](#api-resources)
+    - [Controllers](#controllers)
+    - [Development Environment](#development-environment)
+  - [Test plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+- [Production Readiness](#production-readiness)
+- [Alternatives](#alternatives)
```
shaneutt, line 21:
```suggestion
```

keps/0002-plugins.md
```diff
@@ -0,0 +1,42 @@
...
+- make it possible to add functionality "hooks" for existing functionality
+- enable aftermarket features and API support that don't necessarily make sense
+  to be available upstream to be added dynamically
+- enable building an enterprise version of the operator without forking
+
+### Goals
+
+- make it possible to add go plugins that will add new API support
+- make it possible to add go plugins that will add new webhook functionality
+- make it possible to add go plugins that will load new controllers
```
mflendrich, line 40: These feel indirect. It's not immediately obvious from this description _why_ we need those in the first place. That is: what business value this enables and why that business value makes sense.

An exercise like "The Five Whys" might help us explore those areas better.

rainest:
tl;dr we should determine what the plugins and non-free features are first, consider whether existing Kubernetes tooling can support them also (e.g. the webhook point) or a more limited but sufficient system will work (the Kustomize example at the end), then work on designing a system to handle them.

- If you wish to add your own webhooks, it doesn't seem clear that you'd need our help to do so: nothing is stopping you from creating your own controller-runtime application that spins up a validation server and adding VAW or MAW manifest directing our CRs or other resources through it. How much easier or harder that would be than a plugin system is an open question without some rough design for our proposed interfaces, but it's not impossible without support on our end.
- To echo Michal's comment, the motivations are quite abstract and don't seem to have a clear driving force behind them. There's no "we want to add something that provides functionality X and it is de-coupled from the main operator codebase, so the main operator codebase will allow you to inject it through interface Y", and we need something concrete like that to ensure we're building the correct interface. We shouldn't simply be saying "plugin systems are inherently extensible, extensibility is inherently good, ergo we should build a plugin system and figure out the extensions later"; we should design the plugin system around known needs and leave room for similar future needs, not design for an unknown need and hope a future actual need can fit into that design.
- Kong's purpose dictates different extensibility use cases than ours. Kong generically handles HTTP, and usages of HTTP are many and varied to the point that it's unlikely a single project can handle all of them out of the box. It needs plugins because you're not going to have support for "in the days of yore, a developer at BadArchitectureDesignChoicesCo decided to write their own DSL for routing and stuffed it into HTTP headers, that monstrosity became mission-critical over time, and it's easier for us to use a plugin to hijack normal HTTP routing in favor that DSL" out of the box. We, however, have only a single Kong implementation to support; we shouldn't need to immediately account for non-standard ones that require additional functionality for features that don't exist in stock Kong.
- "add functionality 'hooks' for existing functionality" is unbounded. Unless we expect that you can override all functionality (we should not), this needs to enumerate what functionality you can modify and how.

As one concrete example, a major limitation we know from Helm is that users will want to modify resources that we generate, but can't, because Helm doesn't provide a defined interface to do so (the post-processing system exists, but is very open-ended and isn't configuration-only). To handle that though, I would not propose a Go plugin system, with the expectation that most users do not want to write custom Go code to manipulate resources (and if they did, they could write a MAW to do it).

Providing a standard, limited interface to load some patch DSL (presumably the [standard Kubernetes patching utility](https://github.com/kubernetes-sigs/kustomize)) snippets and apply them, similar to [Kuma's sidecar patch feature](https://kuma.io/docs/1.7.x/explore/dpp-on-kubernetes/#custom-container-configuration) would provide a better UX for most use cases: you only need to write the changes (you have to do this no matter what, even with a more flexible system) using a standard, documented patch language, not the additional code to watch for resources, update them, handle failures, etc. This doesn't handle the most complex of scenarios ("I want to patch a resource differently depending on the phase of the moon, so I need both sets of changes and code to check the moon phase and choose between them"), but it should handle the ones we've often observed ("I want to [inject a static TopologySpreadConstraints into the existing Pod template](https://github.com/Kong/charts/pull/308)").

scseanchow:
One thing I would like captured/articulated in this KEP. If some plugins are going to be enterprise only (theoretically let's say deploying in hybrid mode), how will we gate access to said feature? I'd like to see us explain some interface/requirement of having the user input their enterprise license key to allow for access to 'enterprise' features.

keps/0002-plugins.md
shaneutt: I'm honestly confused about this comment. These three things seemed to me to belong so clearly to the stated motivations:

```
- make it possible to add functionality "hooks" for existing functionality
- enable aftermarket features and API support that don't necessarily make sense
  to be available upstream to be added dynamically
- enable building an enterprise version of the operator without forking
```

You said "these" can you be more specific? Or do you literally feel it is all of them?

shaneutt:
> One thing I would like captured/articulated in this KEP. If some plugins are going to be enterprise only (theoretically let's say deploying in hybrid mode), how will we gate access to said feature? I'd like to see us explain some interface/requirement of having the user input their enterprise license key to allow for access to 'enterprise' features.

Actually I'm trying to avoid any such semantics being in the plugin logic. I've added a `non-goals` section for this. However that's not to say the problem wont be solved: it's simply left up to the plugins _themselves_ to decide how it's solved.

shaneutt:
> tl;dr we should determine what the plugins and non-free features are first, consider whether existing Kubernetes tooling can support them also (e.g. the webhook point) or a more limited but sufficient system will work (the Kustomize example at the end), then work on designing a system to handle them.
>     * If you wish to add your own webhooks, it doesn't seem clear that you'd need our help to do so: nothing is stopping you from creating your own controller-runtime application that spins up a validation server and adding VAW or MAW manifest directing our CRs or other resources through it. How much easier or harder that would be than a plugin system is an open question without some rough design for our proposed interfaces, but it's not impossible without support on our end.

I've dropped webhooks for now, as you suggest it's something we don't _need_ to get started and we can assess how much nicer it is to plug into it later.

>     * To echo Michal's comment, the motivations are quite abstract and don't seem to have a clear driving force behind them. There's no "we want to add something that provides functionality X and it is de-coupled from the main operator codebase, so the main operator codebase will allow you to inject it through interface Y", and we need something concrete like that to ensure we're building the correct interface. We shouldn't simply be saying "plugin systems are inherently extensible, extensibility is inherently good, ergo we should build a plugin system and figure out the extensions later"; we should design the plugin system around known needs and leave room for similar future needs, not design for an unknown need and hope a future actual need can fit into that design.

I've done a re-working of the motivation and goals to help provide clarity, and in the case of goals actually the opposite: make it more generic to make sure we're on the same page about the high level details only.

>     * Kong's purpose dictates different extensibility use cases than ours. Kong generically handles HTTP, and usages of HTTP are many and varied to the point that it's unlikely a single project can handle all of them out of the box. It needs plugins because you're not going to have support for "in the days of yore, a developer at BadArchitectureDesignChoicesCo decided to write their own DSL for routing and stuffed it into HTTP headers, that monstrosity became mission-critical over time, and it's easier for us to use a plugin to hijack normal HTTP routing in favor that DSL" out of the box. We, however, have only a single Kong implementation to support; we shouldn't need to immediately account for non-standard ones that require additional functionality for features that don't exist in stock Kong.

So you disagree with the fundamental motivation of providing an extensibility sandbox, it sounds like. Not sure if I can do anything with that, that would just sound like you're ready to outright block this PR. If that's the case, I would ask that to be productive you please provide alternative solutions for the motivations.
 
>     * "add functionality 'hooks' for existing functionality" is unbounded. Unless we expect that you can override all functionality (we should not), this needs to enumerate what functionality you can modify and how.

Reworded, gave a specific example.

> As one concrete example, a major limitation we know from Helm is that users will want to modify resources that we generate, but can't, because Helm doesn't provide a defined interface to do so (the post-processing system exists, but is very open-ended and isn't configuration-only). To handle that though, I would not propose a Go plugin system, with the expectation that most users do not want to write custom Go code to manipulate resources (and if they did, they could write a MAW to do it).
> Providing a standard, limited interface to load some patch DSL (presumably the [standard Kubernetes patching utility](https://github.com/kubernetes-sigs/kustomize)) snippets and apply them, similar to [Kuma's sidecar patch feature](https://kuma.io/docs/1.7.x/explore/dpp-on-kubernetes/#custom-container-configuration) would provide a better UX for most use cases: you only need to write the changes (you have to do this no matter what, even with a more flexible system) using a standard, documented patch language, not the additional code to watch for resources, update them, handle failures, etc. This doesn't handle the most complex of scenarios ("I want to patch a resource differently depending on the phase of the moon, so I need both sets of changes and code to check the moon phase and choose between them"), but it should handle the ones we've often observed ("I want to [inject a static TopologySpreadConstraints into the existing Pod template](https://github.com/Kong/charts/pull/308)").

This sounds like a counter-proposal, but I don't see it as solving the same problems. For instance, with that kind of a patching strategy how do you propose to load controllers without having to have different managers for them?

shaneutt:
@mflendrich @scseanchow @rainest things are up for review again.

@rainest I found your review in particular hard to be sure what direction you're advocating for. It's very unclear to me whether you're suggesting to stop this effort entirely and take an alternative approach, or if you're just saying you'd like to see the scope significantly reduced. If the latter is the case I would invite you to co-author this KEP with me: please feel free to make updates directly to the language and contents and we can discuss the prescription. Or if it's the former then I would ask for a counter-proposal in the form of a KEP?

shaneutt:
Closing this for now, as there seems to be a strong hunger for heavy deliberation and alternatives. Will potentially revisit after some time synchronizing in zooms and other forums.

PR #93, opened 2022-07-19
ci: Red Hat Scan registry nightly builds

jrsmroz:

Adds nightly builds, that will push images to RedHat scan registry.

**Which issue this PR fixes**

#57

Screenshot 2022-07-26 at 08 43 17

**PR Readiness Checklist**:


.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,133 @@
...
+          tags: ${{ env.REDHAT_STANDARD }}
+
+      - name: Build binary
+        id: docker_build_binary
+        uses: docker/build-push-action@v3
+        with:
+          push: false
+          file: Dockerfile
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache
```
mflendrich, line 76: This may be very ignorant, so please behold:

Does this setup ensure that layers are cached across subsequent CI runs?
This workflow uses `type=local` and no [actions/cache](https://github.com/Kong/kubernetes-ingress-controller/blob/e6c2c585252d5d569e4fa45930ac9701f216c2ff/.github/workflows/test.yaml#L25) to preserve the cache across runs.

There are two ways known to me to achieve this:
1. Use `actions/cache` to load & preserve the state of the layer cache,
2. Use [`type=gha`](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#cache-backend-api) (which I haven't tried myself before)

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,133 @@
...
+jobs:
+  build-push-images:
+    environment: 'Docker Push'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add standard tags
+        run: |
+          echo 'TAGS_STANDARD<<EOF' >> $GITHUB_ENV
+          echo 'type=raw,value=nightly' >> $GITHUB_ENV
+          echo "type=raw,value={{date 'YYYY-MM-DD'}}" >> $GITHUB_ENV
```
mflendrich, line 23: DRY: if we make `{{date 'YYYY-MM-DD'}}` a new variable in `env`, there will be no need for repeated template expansion.

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,133 @@
...
+        uses: docker/login-action@v2.0.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Login to RH Scan Registry 
+        uses: docker/login-action@v2
+        with:
+          registry: scan.connect.redhat.com
```
mflendrich, line 49: Already available in the `RH_SCAN_REGISTRY` variable

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,133 @@
...
+        with:
+          image: quay.io/opdev/preflight:stable
+          options: -v ${{ env.DOCKER_HOME }}:/docker
+          run: |
+            IMAGE_TAG=$(echo ${{ steps.meta_redhat.outputs.tags }} | head -1)
+            /usr/local/bin/preflight check container $IMAGE_TAG \
+              --docker-config=/docker/config.json \
+              --submit \
+              --pyxis-api-token=${{ secrets.RH_PYXIS_TOKEN }} \
+              --certification-project-id=${{ secrets.RH_PROJECT_ID }}
```
mflendrich, line 133: Don't we want to run this before pushing? :thinking:

jrsmroz: No. It requires image to be pushed

.github/workflows/nightly.yaml
jrsmroz: Sure. It was copied from KIC.

.github/workflows/nightly.yaml
shaneutt: It would appear the purpose of this cache is only for between tasks in this job, see the below build entries where the cache is pulled with `cache-from`.

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,133 @@
...
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          file: Dockerfile
+          tags: ${{ steps.meta_redhat.outputs.tags }}
+          cache-from: type=local,src=/tmp/.buildx-cache
+          target: redhat
+          # RedHat scan registry does not support multiarch images
+          # platforms: linux/amd64, linux/arm64
+          platforms: linux/amd64
```
shaneutt, line 112: Right this is what we discussed yesterday. I don't think we need to leave the comment with the multiple platforms in:

```suggestion
          # RedHat scan registry does not support multiarch images
          platforms: linux/amd64
```

However, what I DO think we should add here is a link to the relevant Redhat documentation which describes this limitation, that way future readers have some breadcrumbs to work with if they're trying to make changes to the platform support for this workflow.

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,133 @@
...
+            REPO_INFO=https://github.com/${{ github.repository }}.git
+
+      - name: Setup DOCKER_HOME for preflight
+        run: |
+          echo "DOCKER_HOME=$HOME/.docker/" >> $GITHUB_ENV
+
+      - name: Run preflight
+        uses: addnab/docker-run-action@v3
+        with:
+          image: quay.io/opdev/preflight:stable
```
shaneutt, line 125: We're giving this third party privileged READ access to our Github secrets. I'm not familiar with this action but given that it's not a github or docker action, I would suggest it has greater than usual supply chain security implications. Ideally when using something like this we would separate it out into its own job and give it no secrets access, but that doesn't appear to be completely feasible given that we need to provide it with `RH_PYXIS_TOKEN`.

So my question is: is there any way we can _not_ use this action to fulfill the need?

Dockerfile
```diff
@@ -13,7 +23,11 @@ COPY controllers/ controllers/
 COPY pkg/ pkg/
 COPY internal/ internal/
 
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
+### Distroless/default
+# Use distroless as minimal base image to package the operator binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
```
shaneutt, line 28: Thanks for adding a link to distroless here, I think that will be helpful for future readers :+1:

shaneutt (changes requested):

.github/workflows/nightly.yaml
jrsmroz: There is no doc entry for that. I've learned by experimenting and googling. scan registry refused to accept the multiarch image with internal server error 500, with no additional comments. By googlong I've learned that it's common problem, not all image registries support multiarch yet.

.github/workflows/nightly.yaml
jrsmroz: Can be done. We can create a separate that uses the docker image provided. We do not need to use this action.

.github/workflows/nightly.yaml
jrsmroz: As Shane said.

.github/workflows/nightly.yaml
shaneutt: Well, I can't say I'm particularly surprised that they don't document this, it just fails. Ok then :shrug:

.github/workflows/nightly.yaml
jrsmroz: Can be done. We can create a separate that uses the docker image provided. We do not need to use this action. I need to be separate job thought, not a step in current job.

shaneutt: Unless we have a strong reason then, I suggest we change this.

jrsmroz: Sure. I will try custom action first. Then we can move it separate repo. The KIC will need it too.

.github/workflows/nightly.yaml
rainest: Yeah, the goal here (in the original KIC version, at least) is that we only have to compile the binary once in the builder target, which is then re-used as the basis of both the distroless and UBI images. That generally always needs to change, since the goal is to test the new code that's been merged since the last one.

You could maybe optimize this a bit further to cache other parts across runs, but:
- You'd need to optimize the Dockerfiles first so that any static layers are done before anything that changes run to run.
- Your run times are still going to be dominated by the compile time.
 
There isn't actually much to optimize away, since basically all we do is compile, copy the binary in, and add a user. It'd be of more use if we did something slowish that may well persist between runs, like installing packages.

.github/workflows/nightly.yaml
jrsmroz: @mflendrich fun fact, it's not really a date, it's a format string for and action: https://github.com/docker/metadata-action#date-format

.github/workflows/nightly.yaml
jrsmroz: @shaneutt With github action it looks like this: https://github.com/Kong/gateway-operator/pull/107/files (PR againts this PR)

Are you happy with that? If so we can export that action to separate kong repo and use it for KIC too.

.github/workflows/nightly.yaml
jrsmroz: Acutually there was one bit missing:

```
      - name: Cache Docker layers
        uses: actions/cache@v3
        with:
          path: /tmp/.buildx-cache
          key: ${{ runner.os }}-buildx-${{ github.sha }}
          restore-keys: |
            ${{ runner.os }}-buildx-
```

.github/workflows/nightly.yaml
shaneutt: Yes I think that's good. Thanks for taking the extra time to do that :bow: 

I would like to move forward with that, but we can start by just plugging it in here.

jrsmroz: Yeah. I will extract it to separate repo, so it can be reused, when I go back the KIC RH Certification.

shaneutt (approved):

Issue #65, opened 2022-07-07
Special-case "Immutable" EnvVars for ControlPlanes


shaneutt:
### Problem Statement

Currently we allow end-users to specify _all_ env vars for the control plane, but there are some values that should not be usable when defining the SPEC:

- `CONTROLLER_PUBLISH_SERVICE`
- `CONTROLLER_KONG_ADMIN_URL`
- `CONTROLLER_LEADER_ELECT`
- `CONTROLLER_KONG_ADMIN_TLS_CLIENT_CERT`
- `CONTROLLER_KONG_ADMIN_TLS_CLIENT_CERT_FILE`
- `CONTROLLER_KONG_ADMIN_TLS_CLIENT_KEY`
- `CONTROLLER_KONG_ADMIN_TLS_CLIENT_KEY_FILE`

This list is non-exhaustive, but the commonality is that these are all values that the relevant controller is responsible for dynamically determining and the controller's dynamic value is paramount.

### Acceptance Criteria

- [ ] controller-determined ENV vars like `CONTROLLER_PUBLISH_SERVICE` are not allowed for usage by the validating webhook
- [ ] controller-determined ENV vars like `CONTROLLER_PUBLISH_SERVICE` are stripped out of the SPEC if found during reconcilation, and a (validation) error is logged
- [ ] status entries for the controller-determined values are added to indicate that they're not meant to be declared as part of the intended state, and are instead part of the observed state after automation provisions them

### Additional information

Determine the exhaustive list of options first, and then implement the validation and status fields for them.

rainest:
Continuing from https://github.com/Kong/gateway-operator/pull/103#discussion_r929171841, we shouldn't do this. Users blindly overriding automatically-set variables without documentation instructing them to do so or a known reason that they do need to override is not a common occurrence, and while not unheard of, the fix is a quick "don't do that" support interaction.

We can't know every possible reason someone may want to override a value, but there are definitely cases where doing so makes sense, so we should keep the flexibility to allow them without code changes.

shaneutt:
> Continuing from [#103 (comment)](https://github.com/Kong/gateway-operator/pull/103#discussion_r929171841), we shouldn't do this. Users blindly overriding automatically-set variables without documentation instructing them to do so or a known reason that they do need to override is not a common occurrence, and while not unheard of, the fix is a quick "don't do that" support interaction.
> 
[...]

PR #103, opened 2022-07-22
feat: provision and use mTLS credentials

### LINKED ISSUE #7: Automated Gateway mTLS
### Problem Statement

The `ControlPlane` resource is capable of communicating with the exposed admin api of the `DataPlane` using [mutual TLS (mTLS)](https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/). The purpose of this task is to add automation that creates and manages this mTLS configuration for new instances of the `DataPlane`, and ensures that instances of `ControlPlanes` attached to those `DataPlane` resources can be consumers of those `Secrets` and communicate with the `DataPlane` via the underlying certificates.

### Acceptance Criteria

- [ ] when a `DataPlane` resource is created, mTLS credentials are generated for it by the dataplane-controller
- [ ] any `Deployments` created for a `DataPlane` resource are automatically configured with the mTLS credentials on their admin API listen server
- [ ] when a `ControlPlane` becomes attached to a `DataPlane`, it is configured with the mTLS credentials needed to securely authenticate with the admin API
- [ ] when a `ControlPlane` becomes detached from a `DataPlane` the mTLS credentials it had previously are dropped
- [ ] when a `DataPlane` resource is deleted, it's mTLS `Secrets` are deleted

### Additional Notes

- [ ] TODO comments which linked to this issue are removed once resolved

### LINKED ISSUE #115: Test failure: TestIngressEssentials test flakes

### Current Behavior

TestIngressEssentials is not concurrent modifications safe

```
2022-07-29T13:10:11.9116571Z     ingress_test.go:209: removing the ingress.class annotation "kong" from ingress
2022-07-29T13:10:11.9235351Z     ingress_test.go:216: 
2022-07-29T13:10:11.9237004Z         	Error Trace:	/home/runner/work/gateway-operator/gateway-operator/test/integration/ingress_test.go:216
2022-07-29T13:10:11.9237856Z         	Error:      	Received unexpected error:
2022-07-29T13:10:11.9242137Z         	            	Operation cannot be fulfilled on ingresses.networking.k8s.io "httpbin": the object has been modified; please apply your changes to the latest version and try again
2022-07-29T13:10:11.9243082Z         	Test:       	TestIngressEssentials
2022-07-29T13:10:11.9399706Z --- FAIL: TestIngressEssentials (42.20s)
```

https://github.com/Kong/gateway-operator/runs/7579270802?check_suite_focus=true

### Expected Behavior

Test should not be flaky.

### Steps To Reproduce

_No response_

### Kong Ingress Controller version

_No response_

### Kubernetes version

_No response_

### Anything else?

Proposed solution is to make the code
```go
	switch obj := ingress.(type) {
	case *netv1.Ingress:
		ingress, err := env.Cluster().Client().NetworkingV1().Ingresses(namespace.Name).Get(ctx, obj.Name, metav1.GetOptions{})
		require.NoError(t, err)
		delete(ingress.ObjectMeta.Annotations, annotations.IngressClassKey)
		_, err = env.Cluster().Client().NetworkingV1().Ingresses(namespace.Name).Update(ctx, ingress, metav1.UpdateOptions{})
		require.NoError(t, err)
	case *netv1beta1.Ingress:
		ingress, err := env.Cluster().Client().NetworkingV1beta1().Ingresses(namespace.Name).Get(ctx, obj.Name, metav1.GetOptions{})
		require.NoError(t, err)
		delete(ingress.ObjectMeta.Annotations, annotations.IngressClassKey)
		_, err = env.Cluster().Client().NetworkingV1beta1().Ingresses(namespace.Name).Update(ctx, ingress, metav1.UpdateOptions{})
		require.NoError(t, err)
	}
```

run in `require.Eventually`, check for concurrent modification errors and return false if one is observed.

rainest:
Adds mTLS authentication between ControlPlanes and DataPlanes using the functionality provided by https://github.com/Kong/kubernetes-ingress-controller/pull/2148

**Which issue this PR fixes**
fix #7 
fix #115 

- ~The subject for the ControlPlane certificate is an arbitrary example. Is there something we _should_ be using here? All the nginx side will care about is that it's signed by the CA; it has no expectation for the subject.~ The control plane subject is now `<controlplane.Name>.<controlplane.Namespace>` as a less placeholder arbitrary subject.
- This lacks any handling of certificate rotation. This seems like something we'll need but that is large enough to require its own spike about exactly how to handle it: https://github.com/Kong/gateway-operator/issues/108
- With the expectation that DataPlanes and ControlPlanes can reside in any namespace, and that we need to check for existing certificates, this requires a cluster-wide Secret Get permission. Is there any way we can avoid this?
- ~The user-provided cluster CA Secret must live in the operator's namespace, which is hard-coded right now. Do we have any existing tooling to retrieve this (our own or something in controller-runtime)? I've hardcoded `kong-system` for now.~ I added a downward API envvar to the operator Deployment and access it directly to determine the CA Secret namespace.
- Although the original issue includes it in requirements, this does not implement deletion because there's currently no deletion handling in the reconcilers at all. Why was that missing? Is cleanup in general tracked elsewhere?

**PR Readiness Checklist**:


controllers/controlplane_controller_utils.go
```diff
@@ -49,6 +48,15 @@ func setControlPlaneDefaults(spec *operatorv1alpha1.ControlPlaneDeploymentOption
...
 	}
+	if _, isOverrideDisabled := dontOverride["CONTROLLER_KONG_ADMIN_TLS_CLIENT_CERT_FILE"]; !isOverrideDisabled {
+		spec.Env = updateEnv(spec.Env, "CONTROLLER_KONG_ADMIN_TLS_CLIENT_CERT_FILE", "/var/cluster-certificate/tls.crt")
+	}
+	if _, isOverrideDisabled := dontOverride["CONTROLLER_KONG_ADMIN_TLS_CLIENT_KEY_FILE"]; !isOverrideDisabled {
+		spec.Env = updateEnv(spec.Env, "CONTROLLER_KONG_ADMIN_TLS_CLIENT_KEY_FILE", "/var/cluster-certificate/tls.key")
+	}
+	if _, isOverrideDisabled := dontOverride["CONTROLLER_KONG_ADMIN_CA_CERT_FILE"]; !isOverrideDisabled {
+		spec.Env = updateEnv(spec.Env, "CONTROLLER_KONG_ADMIN_CA_CERT_FILE", "/var/cluster-certificate/ca.crt")
+	}
```
jrsmroz, line 59: Should anyone be able to overwrite those env vars?  We had some discussion with @shaneutt , and basically how we handle env vars in operator should change. Some vars should be setable only by an operator.

shaneutt: That is covered by https://github.com/Kong/gateway-operator/issues/65, I do think it's OK for this PR to continue doing things the way they've been done, as long as we reflect the need to update the things added here in https://github.com/Kong/gateway-operator/issues/65 as well so they get resolved.

rainest: Allowing user overrides of anything is generally good: we can't reliably know every reason someone might want to override our default behavior (e.g. there's a decent case for allowing overrides here to let users issue certs from their own CA infrastructure, since they may want to enforce controls our built-in CA doesn't expose) and allowing arbitrary overrides lets them do so with a simple UX (you just set the variable) and without any code change on our part.

This approach has worked well in the chart. It's quite rare for users to be over-eager and break installs by setting environment variables they think they need, unaware of the automation that would correctly set it.

I do think we need to [change the form of this](https://github.com/Kong/gateway-operator/issues/46), since the current code for checking overrides is awkward, but we should keep the ability for overrides even if we don't think users _should_ need them.

rainest:
Nothing much on the draft version, so handled some of the simpler items and marking it ready to try and force the others.

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -248,3 +249,18 @@ func (r *ControlPlaneReconciler) ensureClusterRoleBindingForControlPlane(
...
+func (r *ControlPlaneReconciler) ensureCertificate(
+	ctx context.Context,
+	controlplane *operatorv1alpha1.ControlPlane,
+) (bool, string, error) {
+	secretName := controlplane.Name + "-control-mtls-cert"
+	usages := []certificatesv1beta1.KeyUsage{certificatesv1beta1.UsageKeyEncipherment,
+		certificatesv1beta1.UsageDigitalSignature, certificatesv1beta1.UsageClientAuth}
+	// this subject is arbitrary. data planes only care that client certificates are signed by the trusted CA, and will
+	// accept a certificate with any subject
+	created, err := maybeCreateCertificateSecret(ctx, fmt.Sprintf("%s.%s", controlplane.Name, controlplane.Namespace),
```
shaneutt, line 262: nit: "maybe" do something can be a bit vague, but if what we're really doing is either GET or PUT, perhaps:

```suggestion
	created, err := getOrCreateCertificateSecret(ctx, fmt.Sprintf("%s.%s", controlplane.Name, controlplane.Namespace),
```

controllers/dataplane_controller_reconciler_utils.go
```diff
@@ -101,9 +102,24 @@ func (r *DataPlaneReconciler) ensureDataPlaneIsMarkedNotProvisioned(
...
 
+func (r *DataPlaneReconciler) ensureCertificate(
+	ctx context.Context,
+	dataplane *operatorv1alpha1.DataPlane,
+	serviceName string,
+) (bool, string, error) {
+	secretName := dataplane.Name + "-data-mtls-cert"
+	usages := []certificatesv1beta1.KeyUsage{certificatesv1beta1.UsageKeyEncipherment,
+		certificatesv1beta1.UsageDigitalSignature, certificatesv1beta1.UsageServerAuth}
+	created, err := maybeCreateCertificateSecret(ctx, fmt.Sprintf("%s.%s.svc", serviceName, dataplane.Namespace),
```
shaneutt, line 113:
```suggestion
	created, err := getOrCreateCertificateSecret(ctx, fmt.Sprintf("%s.%s.svc", serviceName, dataplane.Namespace),
```

go.mod
```diff
@@ -17,21 +17,87 @@ require (
...
 	golang.org/x/crypto v0.0.0-20220411220226-7b82a4e95df4 // indirect
+	golang.org/x/mod v0.6.0-dev.0.20220106191415-9b9b3d81d5e3 // indirect
+	golang.org/x/tools v0.1.10-0.20220218145154-897bd77cd717 // indirect
+	golang.org/x/xerrors v0.0.0-20220609144429-65e65417b02f // indirect
+	google.golang.org/genproto v0.0.0-20220706132729-d86698d07c53 // indirect
+	google.golang.org/grpc v1.47.0 // indirect
+	gopkg.in/cheggaaa/pb.v1 v1.0.28 // indirect
+	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 )
```
shaneutt, line 103: wow that's a lot of dependencies added.

main.go
```diff
@@ -62,10 +64,11 @@ func main() {
...
 	cfg := manager.Config{
-		MetricsAddr:    metricsAddr,
-		ProbeAddr:      probeAddr,
-		LeaderElection: leaderElection,
-		ControllerName: controllerName,
+		MetricsAddr:     metricsAddr,
+		ProbeAddr:       probeAddr,
+		LeaderElection:  leaderElection,
+		ControllerName:  controllerName,
+		ClusterCASecret: clusterCASecret,
```
shaneutt, line 71: So if I'm reading this right, we're effectively _requiring_ that the end-user provides the CA certificate for mTLS connectivity. The intention (if not the letter) of https://github.com/Kong/gateway-operator/issues/7 was to fill in this blank if left empty: it should be the case that a default deployment of the Gateway Operator is capable of setting up mTLS between `ControlPlane` and `DataPlane` objects _all by itself_. So I have a couple questions:

- am I understanding correctly that we don't have this yet, or did I miss something?
- if I am understanding correctly, doesn't this result in the currently `kubectl apply` method of deployment in the README resulting in an operator that doesn't function properly?

shaneutt (changes requested):
The vast majority of this looks great, the code is very clear and understandable and most of my comments are nitpicks. I do however have a larger concern about CA management I would like to make sure we discuss before we merge this.

controllers/controlplane_controller_reconciler_utils.go
rainest: We've used it elsewhere, e.g. https://github.com/Kong/kubernetes-ingress-controller/blob/4a7d27aa4be6bde04b916197b9bdefdc034231f2/internal/manager/controllerdef.go#L48, I think Michal had suggested the convention for functions that may not take action. It's not actually getting anything, just indicating if it did create the certificate, so maybe sounds more appropriate, WDYT?

shaneutt: Sure, and good point. Not a blocker.

controllers/dataplane_controller_reconciler_utils.go
shaneutt: Not a blocker, resolve at your discretion.

main.go
rainest: #7 doesn't say anything about prereqs or lack thereof :smirk_cat: only that the operator provision and use certificates on creation of ControlPlane and DataPlane! https://github.com/Kong/gateway-operator/pull/103/commits/8489bd5707c6c9489a6dddb3167679a09eb39c2f

Using a Runnable kinda obscures failure handling, and I'm not sure there's a way to test that (or how much we care to), but, for example, changing `maybeCreateCACertificate()` to try and use a non-existent namespace will result in

```
INFO: starting the operator's controller manager
INFO: controller output is being logged to /tmp/gateway-operator-controller-logs1692886461
ERROR: problem running manager: namespaces "dgdgfgd" not found
```

internal/manager/run.go
```diff
@@ -179,3 +206,85 @@ func runWebhookServer(mgr manager.Manager, cfg Config) {
...
+
+func (m *caManager) maybeCreateCACertificate() error {
+	// TODO https://github.com/Kong/gateway-operator/issues/108 this also needs to check if the CA is expired and
+	// managed, and needs to reissue it (and all issued certificates) if so
+	ca := &corev1.Secret{}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+	err := m.client.Get(ctx, client.ObjectKey{Namespace: m.secretNamespace, Name: m.secretName}, ca)
+	if errors.IsNotFound(err) {
+		serial, err := rand.Int(rand.Reader, big.NewInt(9999999999))
```
shaneutt, line 228: You could avoid the arbitrary literal large number here and just go with the biggest possible maximum value using constants:

```suggestion
		serial, err := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
```

internal/manager/run.go
```diff
@@ -179,3 +206,85 @@ func runWebhookServer(mgr manager.Manager, cfg Config) {
...
+			Subject: pkix.Name{
+				CommonName:   "Kong Gateway Operator CA",
+				Organization: []string{"Kong, Inc."},
+				Country:      []string{"US"},
+			},
+			SerialNumber:       serial,
+			SignatureAlgorithm: x509.ECDSAWithSHA256,
+			NotBefore:          time.Now(),
+			NotAfter:           time.Now().Add(time.Second * 315400000),
+			//DNSNames:           []string{subject},
```
shaneutt, line 243: We should resolve this comment.

test/integration/suite_test.go
```diff
@@ -142,6 +141,32 @@ func TestMain(m *testing.M) {
 	fmt.Println("INFO: starting the operator's controller manager")
 	go startControllerManager()
 
+	for i := 0; i < 60; i++ {
```
shaneutt, line 144: I'm not necessarily opposed to this approach, but you could also using a timeout, e.g.:

```golang
timeout := time.Now().Add(time.Minute)
for timeout.After(time.Now()) {
```

This has the benefit of the amount of time you sleep between checks no longer having any bearing on the overall time that is allowed, allowing you to use a sleep as little as `time.Millisecond * 50` or none at all perhaps. A micro-optimization for sure, but just something for you to consider while you're already in here.

shaneutt (commented):
Thank you for adding the automated CA setup. Overall I think this is looking good. I have mostly minor comments.

internal/manager/run.go
rainest: Thanks, I was looking for that constant in big and concluded that it didn't exist since there was none there.

internal/manager/run.go
rainest: Yeah, this was a leftover from copying another template and being unsure if it'd trip the SAN requirement (it doesn't, that's only if the CN is a hostname IIRC).

rainest:
I'm taking the :heavy_check_mark: as implicit approval of handling rotation in a separate PR. @shaneutt did you have thoughts on the other two open items in the OP, i.e. handling deletion and the additional Secret permissions?

shaneutt:
> I'm taking the heavy_check_mark as implicit approval of handling rotation in a separate PR. @shaneutt did you have thoughts on the other two open items in the OP, i.e. handling deletion and the additional Secret permissions?

I'm OK with rotation and deletion being handled as separate issues. I might need clarification about the permissions part, if we just mean the permissions needed to manage secrets in this way at all we need those immediately in this PR or a user will have to manually modify the manifests for the operator to work at all.

rainest:
>  I might need clarification about the permissions part, if we just mean the permissions needed to manage secrets in this way at all we need those immediately in this PR or a user will have to manually modify the manifests for the operator to work at all.

We don't currently have permissions to Secrets. To check if we already have certificates in ControlPlane/DataPlane namespaces (which can be any namespace), we need to add Get permissions, granting access to any Secret. This PR includes modifications that add a role that grants access to all Secrets in all namespaces.

This is a sensitive permission that we'd like to avoid if possible, to decrease attack surface for Secret leaks and ease installation if you don't have cluster-wide Secret permissions yourself. I don't think we can really do that without restricting the namespaces that our Deployments can live in.

Restricting namespaces may be viable (offhand I don't know of a reason they'd need to share a namespace with an application we don't already account for), but absent a single blessed namepace (which would pose problems for multi-tenant installs that shouldn't have access to other tenants' Secrets), isn't something we can immediately do with the existing manifests. You'd need something like the --watch-namespaces+chart role automation or manual kustomize work to update namespaces and duplicate the role.

shaneutt:
Ok, then I'm :heavy_check_mark: once we have a follow-up issue in place to further discuss and try to reduce the scope of `Secrets` privileges.

rainest:
Alright, https://github.com/Kong/gateway-operator/issues/121, and deletion?

Unrelated to that, there's apparently some dependency nonsense I need to sort out? Attempting to merge go.sum by stripping merge annotations and tidying resulted in:

```
github.com/kong/gateway-operator/controllers imports
	github.com/cloudflare/cfssl/signer/local imports
	github.com/google/certificate-transparency-go imports
	go.etcd.io/etcd/v3 imports
	go.etcd.io/etcd/tests/v3/integration imports
	go.etcd.io/etcd/pkg/v3/logutil: module go.etcd.io/etcd/pkg/v3@latest found (v3.5.4), but does not contain package go.etcd.io/etcd/pkg/v3/logutil
```

rainest:
Go had already updated `go.etcd.io/etcd/pkg/v3` and but had not updated and was still trying to use versions of `go.etcd.io/etcd/tests/v3/integration` and such. Getting those outdated packages manually fixed it.

No luck on getting the deletes question answered, it seems, so just gonna mark this ready to merge and figure that out later :shrug:

shaneutt (dismissed):

shaneutt (approved):

PR #104, opened 2022-07-25
Add status and shotnames to CRDs

ruromero:
Signed-off-by: ruromero <rromerom@redhat.com>

It adds a bit of simplicity to the use of the Kong Gateway resources by adding short names, groups and additional columns.

* Groups
  * `kubectl get kong` we will get all the dataplane, controlplane and gatewayconfig resources in the namespace
  * `kubectl get all` it will retrieve not only pods, services, configmaps, etc. but also the Kong resources
* Short names
  * `kubectl get kcp` retrieves the Kong ControlPlanes
  * `kubectl get kdp` same for the DataPlanes
  * `kubectl get kgc` retrieves the Kong Gateway Configurations
* Additional Columns: Added the `Ready` and `Provisioned` printer columns for visibility.

### DIFF
--- a/config/crd/bases/gateway-operator.konghq.com_controlplanes.yaml
+++ b/config/crd/bases/gateway-operator.konghq.com_controlplanes.yaml
@@ -13,6 +13,11 @@ spec:
     listKind: ControlPlaneList
     plural: controlplanes
     singular: controlplane
+    shortNames:
+      - kcp
+    categories:
+      - kong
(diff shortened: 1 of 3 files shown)


ruromero:
@shaneutt I don't know if this is something you consider useful.

shaneutt (approved):
No definitely, these are really great UX improvements thank you.

PR #116, opened 2022-07-29
fix: handle webhook availability in test

### LINKED ISSUE #100: Test Failure: TestDataplaneValidatingWebhook
The following error occurred in `main` where the PR that introduced it had succeeded:

```console
INFO: environment is ready, starting tests
=== RUN   TestDataplaneValidatingWebhook
    webhook_test.go:18: start tests
=== RUN   TestDataplaneValidatingWebhook/validating_ok
    webhook_test.go:65: 
        	Error Trace:	/home/runner/work/gateway-operator/gateway-operator/test/e2e/webhook_test.go:65
        	Error:      	Received unexpected error:
        	            	Internal error occurred: failed calling webhook "gateway-operator-validation.konghq.com": failed to call webhook: Post "[https://gateway-operator-validating-webhook.kong-system.svc:443/validate?timeout=5s](https://gateway-operator-validating-webhook.kong-system.svc/validate?timeout=5s)": dial tcp 10.96.236.51:443: connect: connection refused
        	Test:       	TestDataplaneValidatingWebhook/validating_ok
        	Messages:   	test case validating_ok: should not return error when creating dataplane
=== RUN   TestDataplaneValidatingWebhook/database_postgres_not_supported
    webhook_test.go:68: 
        	Error Trace:	/home/runner/work/gateway-operator/gateway-operator/test/e2e/webhook_test.go:68
        	Error:      	"Internal error occurred: failed calling webhook \"gateway-operator-validation.konghq.com\": failed to call webhook: Post \"[https://gateway-operator-validating-webhook.kong-system.svc:443/validate?timeout=5s\](https://gateway-operator-validating-webhook.kong-system.svc/validate?timeout=5s\)": dial tcp 10.96.236.51:443: connect: connection refused" does not contain "database backend postgres of dataplane not supported currently"
        	Test:       	TestDataplaneValidatingWebhook/database_postgres_not_supported
        	Messages:   	test case database_postgres_not_supported: error message should contain expected content
--- FAIL: TestDataplaneValidatingWebhook (5.21s)
    --- FAIL: TestDataplaneValidatingWebhook/validating_ok (0.01s)
    --- FAIL: TestDataplaneValidatingWebhook/database_postgres_not_supported (0.01s)
FAIL
INFO: cleaning up testing cluster and environment
INFO: restore kustomization file from backup file /tmp/kustomization-yaml-backup2878772834
FAIL	github.com/kong/gateway-operator/test/e2e	109.233s
FAIL
make: *** [Makefile:254: test.e2e] Error 1
Error: Process completed with exit code 2.
```

xref: https://github.com/Kong/gateway-operator/runs/7470065296?check_suite_focus=true

rainest:
Retries webhook test stuff if the webhook isn't up yet.

**Which issue this PR fixes**
fix #100

### DIFF
--- a/test/e2e/webhook_test.go
+++ b/test/e2e/webhook_test.go
@@ -4,8 +4,6 @@
 package e2e
 
 import (
-	"errors"
-	"syscall"
 	"testing"
 	"time"
 
@@ -13,6 +11,7 @@ import (
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	netutil "k8s.io/apimachinery/pkg/util/net"
 
 	operatorv1alpha1 "github.com/kong/gateway-operator/apis/v1alpha1"
 )
@@ -69,7 +68,7 @@ func TestDataplaneValidatingWebhook(t *testing.T) {
 				if tc.errMsg == "" {
 					return err == nil
 				}
-				return !errors.Is(err, syscall.ECONNREFUSED)
+				return !netutil.IsConnectionRefused(err)
 			}, time.Minute*3, time.Second)
 			if tc.errMsg == "" {
 				require.NoErrorf(t, err, "test case %s: should not return error when creating dataplane", tc.name)
--- a/test/e2e/webhook_test.go
+++ b/test/e2e/webhook_test.go
@@ -4,7 +4,10 @@
 package e2e
 
 import (
+	"errors"
+	"syscall"
 	"testing"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
@@ -60,7 +63,14 @@ func TestDataplaneValidatingWebhook(t *testing.T) {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			dataplaneClient := operatorClient.ApisV1alpha1().DataPlanes(testNamespace.Name)
-			_, err := dataplaneClient.Create(ctx, tc.dataplane, metav1.CreateOptions{})
+			var err error
+			require.Eventually(t, func() bool {
+				_, err = dataplaneClient.Create(ctx, tc.dataplane, metav1.CreateOptions{})
+				if tc.errMsg == "" {
+					return err == nil
+				}
+				return !errors.Is(err, syscall.ECONNREFUSED)
+			}, time.Minute*3, time.Second)
 			if tc.errMsg == "" {
 				require.NoErrorf(t, err, "test case %s: should not return error when creating dataplane", tc.name)
 			} else {


test/e2e/webhook_test.go
```diff
@@ -60,7 +62,15 @@ func TestDataplaneValidatingWebhook(t *testing.T) {
...
 		t.Run(tc.name, func(t *testing.T) {
 			dataplaneClient := operatorClient.ApisV1alpha1().DataPlanes(testNamespace.Name)
-			_, err := dataplaneClient.Create(ctx, tc.dataplane, metav1.CreateOptions{})
+			var err error
+			require.Eventually(t, func() bool {
+				_, err = dataplaneClient.Create(ctx, tc.dataplane, metav1.CreateOptions{})
+				if tc.errMsg == "" {
+					return err == nil
+				} else {
+					return !strings.Contains(err.Error(), "connection refused")
```
jrsmroz, line 71: There is no defined error, that we could use with `errors.Is` function instead of doing strings comparison?

rainest: Probably the one there now? I didn't actually go and mock the error condition to try and confirm.

jrsmroz: Maybe this https://pkg.go.dev/k8s.io/apimachinery/pkg/util/net#IsConnectionRefused

test/e2e/webhook_test.go
```diff
@@ -60,7 +63,14 @@ func TestDataplaneValidatingWebhook(t *testing.T) {
...
 		t.Run(tc.name, func(t *testing.T) {
 			dataplaneClient := operatorClient.ApisV1alpha1().DataPlanes(testNamespace.Name)
-			_, err := dataplaneClient.Create(ctx, tc.dataplane, metav1.CreateOptions{})
+			var err error
+			require.Eventually(t, func() bool {
+				_, err = dataplaneClient.Create(ctx, tc.dataplane, metav1.CreateOptions{})
+				if tc.errMsg == "" {
+					return err == nil
+				}
+				return !errors.Is(err, syscall.ECONNREFUSED)
```
shaneutt, line 72: This block suggests that if we get an error from the connection, we will check and as long as it's not `ECONNREFUSED` we'll move forward. This suggests other error conditions could crop up, and it doesn't seem ideal that we pass them all through, right?

jrsmroz: Ithink it's ok considering the lines below:
```go
if tc.errMsg == "" {
				require.NoErrorf(t, err, "test case %s: should not return error when creating dataplane", tc.name)
```

test/e2e/webhook_test.go
```diff
@@ -60,7 +62,14 @@ func TestDataplaneValidatingWebhook(t *testing.T) {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			dataplaneClient := operatorClient.ApisV1alpha1().DataPlanes(testNamespace.Name)
-			_, err := dataplaneClient.Create(ctx, tc.dataplane, metav1.CreateOptions{})
+			var err error
+			require.Eventually(t, func() bool {
+				_, err = dataplaneClient.Create(ctx, tc.dataplane, metav1.CreateOptions{})
+				if tc.errMsg == "" {
+					return err == nil
+				}
```
jrsmroz, line 70: I think this block is not needed. That condition is checked after the Eventually call.

test/e2e/webhook_test.go
rainest: https://github.com/kubernetes/apimachinery/blob/v0.24.3/pkg/util/net/util.go#L50-L56 confirms that syscall error, though I prefer dedicated k8s functions when available so will use that.

Attempted to actually confirm behavior but unfortunately there appears to be some issue with it unwrapping? I tested this by changing webhook config mid-run to use invalid ports, and it still failed/did not recognize the error, see https://gist.github.com/rainest/cfaf2887cc3769ea2e2878e23d067b9e and https://cs.opensource.google/go/go/+/refs/tags/go1.18.5:src/errors/wrap.go;l=91-99

Didn't try to review the error generation path (somewhere in or below the generated clientset) to confirm if/where it's not properly wrapping it, but that's my best guess. Am I missing something else here that'd explain why it's not extracting the underlying error?

jrsmroz: Oh, I know. Status error Wraps error in `Reason`.  But it does not implement `As` interface

test/e2e/webhook_test.go
rainest: Yes, but the block after requires that it be the specific error the test wants; it'll just report those errors same as it was reporting ECONNREFUSED in the original report.

This is the only obvious known problem we have here, so we do want the others to fail, to then determine if it's something we broke or some new transient condition we need to handle.

rainest:
Given the annoyance with the error unwrapping and that it makes sense anyway, I looked at making the Pod not ready unless the webhook is already running, but it looks like we're already doing that? Is there any reason that

https://github.com/Kong/gateway-operator/blob/7ef1b7c9f6cce381e1f2b208e1cbaaf106f221ea/internal/manager/run.go#L173

isn't preventing this from happening at all? We already block tests on the Deployment having at least one available replica, so if it's not actually running the server yet it shouldn't proceed with the test.

Checking it directly with a TCP connection is an option but in E2E we don't have easy access to it; we'd need a way into the cluster network or modify manifests to make it a LoadBalancer.

shaneutt:
> Checking it directly with a TCP connection is an option but in E2E we don't have easy access to it; we'd need a way into the cluster network or modify manifests to make it a LoadBalancer.

This is actually something I had suggested before, a temporary LB to get access from outside. That said, if there's a way we can make the readiness work rather than having to do this that is the ideal solution.

test/e2e/webhook_test.go
rainest: Are we indeed stuck with a string comparison here then? I'd rather make it a non-issue by just ensuring it runs, but https://github.com/Kong/gateway-operator/pull/116#issuecomment-1201798086. String comparison is probably good enough given that it's test code and that I'd be pretty surprised if the ERRCONREFUSED string actually changed.

jrsmroz: No, I believe we can first get `var err webhook.ErrCallingWebhook `, the use `err.Reason` in `!netutil.IsConnectionRefused(err.Reason)`

jrsmroz: Disregards. It's not that simple.

rainest:
I am further convinced that this just shouldn't happen, so rather than proceed here with something that ignores the error or adds further readiness checks on the test side, I want to catch a failure after https://github.com/Kong/gateway-operator/pull/123 to try and determine how the heck we're getting connections refused at this point.

We may find that it's something unavoidable, in which case I plan to proceed with the error string scrape approach. Converting back to a draft until we understand the failure better.

PR #119, opened 2022-08-01
ci: add release pipeline

jrsmroz:

Adds release pipeline, that will publish images to the Red Har scan registry.

**Which issue this PR fixes**

#57 

**PR Readiness Checklist**:


.github/workflows/release.yaml
```diff
@@ -0,0 +1,242 @@
...
+      - name: checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Kubernetes ${{ matrix.kubernetes_version }} Integration Tests
+        run: make test.integration
+
+      - name: Kubernetes ${{ matrix.kubernetes_version }} E2E Tests
+        run: make test.e2e
```
pmalek, line 221: Wouldn't it be beneficial for time efficiency's sake to run 2 jobs separately: 1 for e2e and 1 for integration and require both for publish-release?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,242 @@
...
+      with:
+        body: |
+          #### Download Kong Gateway Operator ${{ needs.semver.outputs.fullversion }}:
+
+          - [Docker Image](https://github.com/Kong/gateway-operator/pkgs/container/gateway-operator)
+          - [Get started](https://github.com/Kong/gateway-operator/blob/main/README.md)
+
+          #### Links:
+
+          - [Changelog](https://github.com/Kong/gateway-operator/blob/main/CHANGELOG.md#${{ needs.semver.outputs.major }}${{ needs.semver.outputs.minor }}${{ needs.semver.outputs.patch }})
```
pmalek, line 238: I don't see the `semver` job having the `patch` output.

.github/workflows/release.yaml
jrsmroz: It make sense, I can change that, thought it's not critical, considering release is not run every time.

rainest (approved):
I also agree that it'd make sense to run E2E in parallel to fail faster if it's going to fail, but it's not critical. Rest looks good.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,243 @@
...
+
+      - name: Kubernetes ${{ matrix.kubernetes_version }} E2E Tests
+        run: make test.e2e
+
+
+  publish-release:
+    runs-on: ubuntu-latest
+    needs: [semver, build-push-images, test-current-kubernetes]
+    steps:
+    - uses: ncipollo/release-action@v1
```
shaneutt, line 229: It seems that we're so commonly finding third party actions which we want to use, but they always give me pause as in a privileged context (such as this where we give that code our `GITHUB_TOKEN`) they represent possible supply chain attack vectors.

The only saving grace I'm aware of here is that I believe the `GITHUB_TOKEN` in these contexts is a limited time window and expires when the run ends, but I would need to verify that by finding the documentation again.

In this case, how problematic is it for us to avoid this dependency so as to not feed it our API token?

shaneutt (changes requested):

.github/workflows/release.yaml
jrsmroz: Well, it's from KIC, so I guess if we want to change that, we should so in KIC too.

jrsmroz:
> I also agree that it'd make sense to run E2E in parallel to fail faster if it's going to fail, but it's not critical. Rest looks good.

Ok, two votes, let me change that than.

.github/workflows/release.yaml
jrsmroz: Done

.github/workflows/release.yaml
```diff
@@ -0,0 +1,268 @@
+name: release
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'The version to release (e.g. v1.2.3)'
```
pmalek, line 7: Nit: in yaml you don't need to quote a string if it's not a number (just as you have it in other places in this file.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,268 @@
...
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Kubernetes ${{ matrix.kubernetes_version }} E2E Tests
+        run: make test.e2e
+
+  publish-release:
+    runs-on: ubuntu-latest
+    needs: [semver, build-push-images, test-integration-current-kubernetes, test-e2e-current-kubernetes]
```
pmalek, line 252: Nit: In other places we're using arrays where each element is on its own line. Why use a different approach here? IMHO it's a bit more readable when each entry has it's own line + it's clearer on the diff when adding/removing elements. WDYT?

.github/workflows/release.yaml
jrsmroz: ok

.github/workflows/release.yaml
jrsmroz: That's true. Fixed.

pmalek (approved):
LGTM 👍

.github/workflows/release.yaml
shaneutt: Given the precedent I wont block. Would you mind creating a issue to follow up on this? If not, I'm happy to make that issue, either way. ty.

shaneutt (approved):

.github/workflows/release.yaml
jrsmroz: I will. Migrating to GitHub owned action seems safer choice. Not sure what was the story behind using ncipollo/release-action@v1

jrsmroz: https://github.com/Kong/gateway-operator/issues/128

PR #123, opened 2022-08-02
chore: add diagnostics for webhook test

rainest:
#100 [doesn't seem like it should happen](https://github.com/Kong/gateway-operator/issues/100#issuecomment-1203153747). It appears something weird is going on and we shouldn't just expect that sometimes the webhook won't be available when needed and check before individual tests a la #116.

This dumps cluster state to hopefully explain how we're getting these errors after the Deployment becomes ready. Depending on what we learn, we may decided to proceed with #116 or we may find something broken in the tests that we should fix differently.

This adds some TODOs where we have duplicate functionality that KTF should handle.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -93,6 +93,14 @@ jobs:
       env:
         KONG_CONTROLLER_OUT: stdout
 
+    - name: upload diagnostics
+      if: ${{ always() }}
+      uses: actions/upload-artifact@v3
+      with:
+        name: diagnostics-integration-tests-enterprise-postgres
+        path: /tmp/ktf-diag*
+        if-no-files-found: ignore
+
   e2e-tests:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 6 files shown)


test/e2e/utils_test.go
```diff
@@ -10,20 +10,29 @@ import (
...
-func createNamespaceForTest(t *testing.T) (*corev1.Namespace, func()) {
-	namespaceForTestCase, err := clusters.GenerateNamespace(ctx, env.Cluster(), t.Name())
-	exitOnErr(err)
-
-	cleanup := func() {
-		assert.NoError(t, clusters.CleanupGeneratedResources(ctx, env.Cluster(), t.Name()))
-	}
-
-	return namespaceForTestCase, cleanup
+// TODO https://github.com/Kong/kubernetes-testing-framework/issues/302
```
shaneutt, line 17: :+1:

shaneutt (approved):

.github/workflows/tests.yaml
```diff
@@ -93,6 +93,14 @@ jobs:
       env:
         KONG_CONTROLLER_OUT: stdout
 
+    - name: upload diagnostics
+      if: ${{ always() }}
+      uses: actions/upload-artifact@v3
+      with:
+        name: diagnostics-integration-tests-enterprise-postgres
```
pmalek, line 100: I wonder why those are named `...-enterprise-postgres` given that enterprise is not mentioned anywhere in this job's definition.

PR #130, opened 2022-08-03
feat: update ControlPlane with new publish Service

### LINKED ISSUE #97: Update dataplane env vars in controlplane when dataplane service changes

### Problem Statement

The `ControlPlane` deployments are pointed to the correct `DataPlane` service via 2 different env vars: `CONTROLLER_PUBLISH_SERVICE`, and `CONTROLLER_KONG_ADMIN_URL`. These 2 vars contain the correct dataplane service address to talk to. With https://github.com/Kong/gateway-operator/pull/91, when the `DataPlane` service is deleted, it is recreated automatically by the `dataplane-controller`; the problem is that the service name is not static, but created through the `generateName` field. Hence, every time the `DataPlane`service is recreated by the `dataplane-controller`, the `controlplane-controller` should also update the `ControlPlane` deployment env vars.

### Proposed Solution

We should trigger an update on the `controlplane-controller` every time the `DataPlane` service changes. To do so, we can watch the `Dataplane` services in the `controlplane-controller` and trigger a reconciliation loop every time an event occurs.

### Additional information

_No response_

### Acceptance Criteria

- [ ] Every time the dataplane service is created, the dataplane service env vars are set in the controlplane deployment
- [ ] Every time the dataplane service is deleted, the dataplane service env vars are unset from the controlplane deployment

rainest:
Lists the current Service for a DataPlane in the DataPlane's status.

Adds a controller-runtime reference between a ControlPlane and the DataPlane it configures. DataPlane updates trigger an update of the ControlPlane.

Overwrite ControlPlane Deployment environments if they differ from the ControlPlane environment on ControlPlane reconciles (we were already updating the ControlPlane environment with the latest DataPlane Service on reconciles).

**Which issue this PR fixes**
Fix #97 

PR to another branch with an open PR. We filed an issue for this, but it doesn't look like it's relevant without that other branch.

Previously, our reconciliation strategy for ControlPlanes meant that you (or some application with access to do it) could modify their Deployment's environment by editing the Deployment itself. With this change, that is no longer possible-ish--we will overwrite edits on DataPlane changes or anything else that triggers a ControlPlane reconcile. This may or may not be fine, but we should be aware of the change at least.

**PR Readiness Checklist**:

- [ ] ~the `CHANGELOG.md` release notes have been updated to reflect significant changes~ ALPHA LIMBO


controllers/controlplane_controller.go
```diff
@@ -86,14 +88,26 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	debug(log, "retrieving dataplane info", controlplane)
+	var dataplaneServiceName string
+	dataplane, err := gatewayutils.GetDataPlaneForControlPlane(ctx, r.Client, controlplane)
 	if err != nil {
 		if !errors.Is(err, operatorerrors.ErrDataPlaneNotSet) {
 			return ctrl.Result{}, err
 		}
 
 		debug(log, "no existing dataplane for controlplane", controlplane, "error", err)
+		dataplaneServiceName = ""
```
jrsmroz, line 100: Hmm, that's not needed.

rainest: Originally I thought this would own/watch the Service, and found that we had no way to retrieve it because the closest utility was returning on the name after retrieving the Service. That ended up not being the right approach (the original Service is getting destroyed and isn't what we should watch as such), but it did seem like we'd want the general-purpose utility function to return the entire object, to let its users choose which Service properties they actually need.

In practice we are only using the name so far, because our current use for it is informing the controller about the Service, and the controller performs its own Service get.

https://github.com/Kong/gateway-operator/pull/130/commits/1f633659f100b3786062fabfa45abdd8e6c0e697 refactors downstreams around using a Service object instead of extracting the name in their caller; we could either go that route to use the Service object or back out the original change `GetDataPlaneService()` back to `GetDataPlaneServiceName()`. WDYT?

controllers/controlplane_controller.go
```diff
@@ -50,6 +51,7 @@ func (r *ControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Named("ControlPlane").
 		Owns(&appsv1.Deployment{}).
 		Owns(&corev1.ServiceAccount{}).
+		Owns(&operatorv1alpha1.DataPlane{}).
```
mlavacca, line 54: I don't think we should use `Owns` here as it is intended to be used only for resource generated and owned by the controller, as it is stated in the godoc:
```
// Owns defines types of Objects being *generated* by the ControllerManagedBy, and configures the ControllerManagedBy to respond to
// create / delete / update events by *reconciling the owner object*.  This is the equivalent of calling
// Watches(&source.Kind{Type: <ForType-forInput>}, &handler.EnqueueRequestForOwner{OwnerType: apiType, IsController: true}).
```

We should use the underlying API instead

mlavacca, line 54: I don't think we should use `Owns` here as it is intended to be used only for resources generated and owned by the controller, as it is stated in the godoc:
```
// Owns defines types of Objects being *generated* by the ControllerManagedBy, and configures the ControllerManagedBy to respond to
// create / delete / update events by *reconciling the owner object*.  This is the equivalent of calling
// Watches(&source.Kind{Type: <ForType-forInput>}, &handler.EnqueueRequestForOwner{OwnerType: apiType, IsController: true}).
```

We should use the raw `watches` method instead.

mlavacca:
Thank you for implementing this @rainest. I know this PR is kinda useless without [this](https://github.com/Kong/gateway-operator/pull/91) merged. I'm going to restart working on that PR and get it merged so that we can unblock this one.

rainest:
Draft pending parent.

rainest:
Superseded by https://github.com/Kong/gateway-operator/pull/218

PR #131, opened 2022-08-04
feat: add telemetry

pmalek:

This adds basic telemetry anonymized reports to gateway operator.

- ~~this code does not compile yet because it relies on unmerged code (hence the `replace` in `go.mod`)~~
- ~~https://github.com/Kong/kubernetes-telemetry/ is private hence this won't compile (on top of the above)~~
- only ping signal is implemented, i.e. first report is sent out **after** the telemetry sending period has passed (for now, after 1h)
  - This is to be implemented via https://github.com/Kong/kubernetes-telemetry/issues/38


Makefile
```diff
@@ -1,3 +1,16 @@
+# ------------------------------------------------------------------------------
+# Configuration - Repository
+# ------------------------------------------------------------------------------
+
+REPO ?= github.com/kong/gateway-operator
+REPO_INFO ?= $(shell git config --get remote.origin.url)
+# TODO: remove --always after first tag
+TAG ?= $(shell git describe --tags --always)
```
rainest, line 8: We should be able to just leave this in its final form, correct? The first proper image we ship should be from a tagged commit, so we don't really need to worry about the current limbo period.
```suggestion
TAG ?= $(shell git describe --tags --always)
```
Worst case, this is just outputting nothing to stdout at present, so any images we do build before then just get an empty tag and show up in stats as an unknown version, which isn't the end of the world (it's arguably correct!).

rainest (changes requested):
Minor thing around a future TODO we can probably avoid, otherwise looks good.

Makefile
pmalek: 😐 well.... allllllllllllright but your suggestion doesn't reflect what I think you wanted to convey :)

rainest (approved):

PR #144, opened 2022-08-10
fix: gc cluster wide resources owned by ControlPlane on delete

### LINKED ISSUE #67: Cluster-wide resources garbage collection

### Problem Statement

With the introduction of #58, the `ClusterRole` and `ClusterRoleBInding` for the controlplane deployments are automatically created. Each of those resources has an OwnerReference to the controlplane to be correctly referenced. We use this approach for all the other resources also to have garbage collection for free (each controlplane deletion implies also the controlplane `deployment` and `ServiceAccount` deletion), but this approach isn't enough for the cluster-wide resources. If a cluster-wide resource is owned by a namespaced resource, the garbage-collection isn't performed, leading to having RBAC leftovers. 

### Proposed Solution

Addition of finalizers on the controlplane to be able to delete `ClusterRoles` and `ClusterRoleBindings` upon controlplane deletion.

randmonkey:

garbage collect cluster wide resources (currently `ClusterRoles`,`ClusterRoleBindings`) owned by controlplanes when they are deleted.

**Which issue this PR fixes**
fixes #67 

**PR Readiness Checklist**:


test/integration/controlplane_test.go
```diff
@@ -191,4 +195,11 @@ func TestControlPlaneEssentials(t *testing.T) {
 	t.Log("verifying custom env TEST_ENV has value configured in controlplane")
 	testEnvValue := getEnvValueByName(envs, "TEST_ENV")
 	require.Equal(t, "test", testEnvValue)
+
+	// delete controlplane and verify that cluster wide resources removed.
+	t.Log("verifying cluster wide resources removed after controlplane deleted")
+	err = controlplaneClient.Delete(ctx, controlplane.Name, metav1.DeleteOptions{})
```
pmalek, line 201: Since we're using `require` above this line and those can fail then this line here can never trigger and we'll never delete the control plane, causing garbage to be left in the cluster.

Can we move this up higher then?

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -267,3 +267,46 @@ func (r *ControlPlaneReconciler) ensureCertificate(
...
+		consts.GatewayOperatorControlledLabel, consts.ControlPlaneManagedLabelValue, controlplane.UID,
+	)
+	if err != nil {
+		return err
+	}
+	for i := range clusterRoleBindings {
+		err = r.Client.Delete(ctx, &clusterRoleBindings[i])
+		if err != nil {
+			return err
+		}
```
pmalek, line 289: I think we don't want to return here. The reason for this is that (if I understand this right) we want to remove as much as possible, not "fail fast" during removal.

So perhaps we can use something like https://github.com/hashicorp/go-multierror or an alternative and just group errors but at the same time continue execution to remove as much as possible?

pmalek (changes requested):

controllers/controlplane_controller_consts.go
```diff
@@ -0,0 +1,13 @@
...
+// -----------------------------------------------------------------------------
+// ControlPlane - Finalizers
+// -----------------------------------------------------------------------------
+
+type ControlPlaneFinalizer string
+
+const (
+	// ControlPlaneFinalizerClusterResourcesGC is the finalizer to garbage collect
+	// cluster wide resources owned by the ControlPlane (e.g: ClusterRoles, ClusterRoleBindings)
+	ControlPlaneFinalizerClusterResourcesGC ControlPlaneFinalizer = "gateway-operator.konghq.com/cluster-resource-gc"
```
mlavacca, line 12: Instead of using a single Omni-comprehensive finalizer, I rather think that using a finalizer for each resource type (e.g., `gateway-operator.konghq.com/cleanup-clusterrole`,  `gateway-operator.konghq.com/cleanup-clusterrolebinding`) is more meaningful for what resources are still to be collected and which are left in the cluster. If we pursue this way, in `garbageCollectClusterResources` we should look up for every finalizer before trying to delete the corresponding resource. WDYT?

mlavacca: Furthermore, don't we have the prefix `gateway-operator.konghq.com` already stored in some variable that we can re-use to build the finalizers?

controllers/controlplane_controller_consts.go
```diff
@@ -0,0 +1,13 @@
+package controllers
+
+// -----------------------------------------------------------------------------
+// ControlPlane - Finalizers
+// -----------------------------------------------------------------------------
+
+type ControlPlaneFinalizer string
```
pmalek, line 7: What is the purpose for this type? While generally I'd advocate for having separate types I don't see a benefit of using it here. Can you help me understand the reason for this?

test/integration/controlplane_test.go
randmonkey: > Can we move this up higher then?
higher then what?

I don't know what is your expectation here. In integration tests, the whole test cluster would be destroyed after the test finished by default, so we do not care about whether garbage are in the cluster after test finish (no matter passed or failed).

controllers/controlplane_controller_consts.go
randmonkey: > Furthermore, don't we have the prefix `gateway-operator.konghq.com` already stored in some variable that we can re-use to build the finalizers?

I did not find this prefix. Should we define it somewhere? as a repo-level constant, or just in `controller` package or defining finalizers ?

test/integration/controlplane_test.go
pmalek: ~~Yes, in CI the cluster will be spawned just once and then destroyed but there might be cases where devs could run just a single test to verify functionality, bug fix, or something else and they'd like to reuse a cluster.~~

Ok, it seems that we already have a cleaner call in line 159: `cleaner.Add(controlplane)` which should cover what I described

controllers/controlplane_controller_consts.go
mlavacca: We already have a file in which to store all consts, under `internal/consts/consts.go`. I think that place could be a good place to put this finalizer, instead of creating this new file.

controllers/controlplane_controller_reconciler_utils.go
randmonkey: updated.

controllers/controlplane_controller_consts.go
randmonkey: > Instead of using a single Omni-comprehensive finalizer, I rather think that using a finalizer for each resource type (e.g., `gateway-operator.konghq.com/cleanup-clusterrole`, `gateway-operator.konghq.com/cleanup-clusterrolebinding`) is more meaningful for what resources are still to be collected and which are left in the cluster. WDYT?

used two finalizers for cleaning up each kind of resource, and moved the removing function to separate functions.

randmonkey: > We already have a file in which to store all consts, under `internal/consts/consts.go`. I think that place could be a good place to put this finalizer, instead of creating this new file.

IMHO, constants in `internal/consts/consts.go` are expected to be used in multiple packages or in different controllers, not a place to put ALL constants used in this repo. These finalizers are used only in controlplane controller, so I made a file to store them.

controllers/controlplane_controller_consts.go
randmonkey: This is used for enumerate possible finalizers added by controplane controller. added comments here.

controllers/controlplane_controller.go
```diff
@@ -48,6 +49,26 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		return ctrl.Result{}, err.ErrorOrNil()
+	}
+
+	// ensure the controlplane has a finalizer to delete owned cluster wide resources on delete.
+	metaChanged := k8sutils.EnsureFinalizerInMetadata(&controlplane.ObjectMeta, string(ControlPlaneFinalizerCleanupClusterRole))
+	metaChanged = metaChanged || k8sutils.EnsureFinalizerInMetadata(&controlplane.ObjectMeta, string(ControlPlaneFinalizerCleanupClusterRoleBinding))
+	if metaChanged {
+		info(log, "update metadata of control plane to set finalizer", controlplane.ObjectMeta)
+		err := r.Client.Update(ctx, controlplane)
+		return ctrl.Result{}, err
```
pmalek, line 69: nitpick: 

```suggestion
		return ctrl.Result{}, r.Client.Update(ctx, controlplane)
```

controllers/controlplane_controller.go
```diff
@@ -48,6 +49,26 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	// controlplane is deleted, just run garbage collection for cluster wide resources.
+	if !controlplane.DeletionTimestamp.IsZero() {
+		debug(log, "remove owned cluster wide resources", controlplane)
+		var err *multierror.Error
+		removeCrbErr := r.removeOwnedClusterRoleBindings(ctx, controlplane)
+		err = multierror.Append(err, removeCrbErr)
+		removeCrErr := r.removeOwnedClusterRoles(ctx, controlplane)
+		err = multierror.Append(err, removeCrErr)
+		return ctrl.Result{}, err.ErrorOrNil()
+	}
```
pmalek, line 61: 👍

internal/utils/kubernetes/metadata.go
```diff
@@ -15,3 +16,39 @@ import (
...
+		}
+	}
+
+	metadata.Finalizers = append(metadata.Finalizers, finalizer)
+	return true
+}
+
+// RemoveFinalizerInMetadata removes the finalizer from the finalizers in ObjectMeta.
+// if it exists, remove the finalizer from the slice.
+// retruns true if the ObjectMeta is really changed.
```
pmalek, line 36:
```suggestion
// If it exists, removes the finalizer from the slice.
// Returns true if the ObjectMeta has been changed.
```

internal/utils/kubernetes/metadata.go
```diff
@@ -15,3 +16,39 @@ import (
 func GetAPIVersionForObject(obj client.Object) string {
 	return fmt.Sprintf("%s/%s", obj.GetObjectKind().GroupVersionKind().Group, obj.GetObjectKind().GroupVersionKind().Version)
 }
+
+// EnsureFinalizerInMetadata ensures the expected finalizer exist in ObjectMeta.
+// if the finalizer does not exist, append it to finalizers.
+// returns true if the ObjectMeta is really changed.
```
pmalek, line 22:
```suggestion
// If the finalizer does not exist, appends it to finalizers.
// Returns true if the ObjectMeta has been changed.
```

pmalek (commented):
Overall 👍 Left some nitpicks

controllers/controlplane_controller.go
randmonkey: fixed.

internal/utils/kubernetes/metadata.go
randmonkey: updated.

internal/utils/kubernetes/metadata.go
randmonkey: updated.

controllers/controlplane_controller_consts.go
```diff
@@ -0,0 +1,15 @@
+package controllers
+
+// -----------------------------------------------------------------------------
+// ControlPlane - Finalizers
+// -----------------------------------------------------------------------------
+
+// ControlPlaneFinalizer are possible finalizers added by controlplane controller.
```
pmalek, line 7: nit: (although I'm not sure this fits 100%)

```suggestion
// ControlPlaneFinalizer defines finalizers added by controlplane controller.
```

pmalek (approved):
Ideally others would also chip in but this LGTM as of now.

controllers/controlplane_controller_consts.go
```diff
@@ -0,0 +1,15 @@
...
+type ControlPlaneFinalizer string
+
+const (
+	// ControlPlaneFinalizerCleanupClusterRole is the finalizer to cleanup clusterroles owned by controlplane on deleting.
+	ControlPlaneFinalizerCleanupClusterRole ControlPlaneFinalizer = "gateway-operator.konghq.com/cleanup-clusterrole"
```
mlavacca, line 12: I personally keep advocating for moving these constants under internal/consts/consts.go. We have the labels to retrieve the objects stored in that file, along with k8s name prefixes and so on. TBH, I think that scattering around constants that are functionally correlated (all of them are meant to enrich K8s objects with additional info) will lead to a code quality decrease.

mlavacca: Other than this aspect, this lgtm

controllers/controlplane_controller.go
```diff
@@ -48,6 +49,25 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
+	// controlplane is deleted, just run garbage collection for cluster wide resources.
+	if !controlplane.DeletionTimestamp.IsZero() {
+		debug(log, "remove owned cluster wide resources", controlplane)
+		var err *multierror.Error
+		removeCrbErr := r.removeOwnedClusterRoleBindings(ctx, controlplane)
+		err = multierror.Append(err, removeCrbErr)
+		removeCrErr := r.removeOwnedClusterRoles(ctx, controlplane)
+		err = multierror.Append(err, removeCrErr)
+		return ctrl.Result{}, err.ErrorOrNil()
```
shaneutt, line 60: Reading this block it's not clear where the finalizers are actually being removed at a glance. You have to dig in to the binding and role removal methods to find out that's where's it's being done. For future code readers it would be better to have upfront clarity on the finalizer handling. Another side effect of how this is currently done is that under some (albeit minor) situations you could end up with the object in the queue multiple times.

Here's an alternative implementation for your considerations:

```go
if !controlplane.DeletionTimestamp.IsZero() {
	// wait for termination grace period before cleaning up roles and bindings
	if controlplane.DeletionTimestamp.After(metav1.Now()) {
		debug(log, "control plane deletion still under grace period", controlplane)
		return ctrl.Result{RequeueAfter: constants.GracePeriodRequeueDelay}
	}

	// ensure that the rolebindings which were created for the ControlPlane are deleted
	if err := r.ensureOwnedClusterRoleBindingsDeleted(ctx, controlplane); err != nil {
		return ctrl.Result{}, err
	}

	// now that RoleBindings are cleaned up, remove the relevant finalizer
	if k8sutils.RemoveFinalizerInMetadata(&controlplane.ObjectMeta, string(ControlPlaneFinalizerCleanupClusterRoleBinding)) {
		return ctrl.Result{}, r.Client.Update(ctx, controlplane) // requeue here until rolebinding finalizer is removed 
	}

	// ensure that the roles created for the controlplane are deleted
	if err := r.ensureOwnedClusterRolesDeleted(ctx, controlplane); err != nil {
		return ctrl.Result{}, err
	}

	// now that Roles are cleaned up, remove the relevant finalizer
	if k8sutils.RemoveFinalizerInMetadata(&controlplane.ObjectMeta, string(ControlPlaneFinalizerCleanupClusterRole)) {
		return ctrl.Result{}, r.Client.Update(ctx, controlplane) // requeue here until role finalizer is removed
	}

	return ctrl.Result{}, nil // cleanup completed
}
```

So with the above example we would end up with more clarity up front on how finalizers are being handled: the code reader can see that we're very intentionally deleting sub-resources and then removing the relevant finalizer at a glance. Also with the above we're expressing the re-entrant code in a way that more clearly defines re-queues as happening immediately after the object is updated which in general can be beneficial for code clarity, reducing problems like extra enqueuing of the same object, and otherwise make the logic more predictable.

All that said, these things are very minor and so I don't consider this comment a blocker. I would just ask that you consider these points and see if there's ways we can improve on readability and re-entry.

controllers/controlplane_controller.go
```diff
@@ -48,6 +49,25 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	}
+
+	// ensure the controlplane has a finalizer to delete owned cluster wide resources on delete.
+	metaChanged := k8sutils.EnsureFinalizerInMetadata(&controlplane.ObjectMeta, string(ControlPlaneFinalizerCleanupClusterRole))
+	metaChanged = metaChanged || k8sutils.EnsureFinalizerInMetadata(&controlplane.ObjectMeta, string(ControlPlaneFinalizerCleanupClusterRoleBinding))
+	if metaChanged {
```
shaneutt, line 66: nit: `metaChanged` is a very general way to describe what the results of the `EnsureFinalizerInMetadata` function does, what do you think about this alternative name?

```suggestion
	finalizersChanged := k8sutils.EnsureFinalizerInMetadata(&controlplane.ObjectMeta, string(ControlPlaneFinalizerCleanupClusterRole))
	finalizersChanged = finalizersChanged || k8sutils.EnsureFinalizerInMetadata(&controlplane.ObjectMeta, string(ControlPlaneFinalizerCleanupClusterRoleBinding))
	if finalizersChanged {
```

shaneutt (commented):
No blockers, but a couple comments for your considerations.

controllers/controlplane_controller.go
```diff
@@ -48,6 +49,25 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 
+	// controlplane is deleted, just run garbage collection for cluster wide resources.
+	if !controlplane.DeletionTimestamp.IsZero() {
```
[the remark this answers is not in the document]
randmonkey: Correct me if I made any mistakes: the deletion timestamp could NOT be unset or set to a later, so we could requeue the controlplane when the grace period expires. If the deletion timestamp is changed, the update would trigger another round of reconclilation.

controllers/controlplane_controller.go
randmonkey:
```go
if controlplane.DeletionTimestamp.After(metav1.Now()) {
		debug(log, "control plane deletion still under grace period", controlplane)
		return ctrl.Result{RequeueAfter: constants.GracePeriodRequeueDelay}
	}
```
Correct me if I made any mistakes: the deletion timestamp could NOT be unset or set to a later, so we could requeue the controlplane when the grace period expires (as in the following snippet) If the deletion timestamp is changed, the update would trigger another round of reconclilation.

```go
		if controlplane.DeletionTimestamp.After(metav1.Now().Time) {
			debug(log, "control plane deletion still under grace period", controlplane)
			return ctrl.Result{
				Requeue: true,
				// Requeue when grace period expires.
				// If deletion timestamp is changed,
				// the update will trigger another round reconilation.
				// so we do not consider updates of deletion timestmp here.
				RequeueAfter: time.Until(controlplane.DeletionTimestamp.Time),
			}, nil
		}
```

controllers/controlplane_controller.go
randmonkey: fixed.

controllers/controlplane_controller.go
randmonkey: taken this implementation.

controllers/controlplane_controller_consts.go
randmonkey: I put these constants in separate file because I took the similar usage of `*_conditions.go`. These are only used in one controller, so they are not put in the common `consts` package. The finalizers are also only used in one controller, I think I need a reason to treat them in different way with the conditions of each resource.

controllers/controlplane_controller_consts.go
randmonkey: updated.

controllers/controlplane_controller_consts.go
mlavacca: well, I actually never said that we should have a large constant pool for ALL the constants of the projects, I was referring only to the ones used to enrich K8s objects, as labels, annotations, or finalizers. Btw, I guess it is fine as it is.

controllers/controlplane_controller.go
shaneutt: Yes this is a good way to do it, I like it :+1:

controllers/controlplane_controller.go
```diff
@@ -48,6 +50,57 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		// wait for termination grace period before cleaning up roles and bindings
+		if controlplane.DeletionTimestamp.After(metav1.Now().Time) {
+			debug(log, "control plane deletion still under grace period", controlplane)
+			return ctrl.Result{
+				Requeue: true,
+				// Requeue when grace period expires.
+				// If deletion timestamp is changed,
+				// the update will trigger another round reconilation.
+				// so we do not consider updates of deletion timestmp here.
```
shaneutt, line 63:
```suggestion
				// so we do not consider updates of deletion timestamp here.
```

controllers/controlplane_controller.go
```diff
@@ -48,6 +50,57 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+				RequeueAfter: time.Until(controlplane.DeletionTimestamp.Time),
+			}, nil
+		}
+
+		debug(log, "remove owned cluster wide resources", controlplane)
```
shaneutt, line 68:
```suggestion
		debug(log, "removing owned roles and role bindings", controlplane)
```

shaneutt (dismissed):
couple more super tiny comments, but then looks really good!

controllers/controlplane_controller.go
randmonkey: fixed typos here.

controllers/controlplane_controller.go
randmonkey: updated.

internal/utils/kubernetes/metadata.go
```diff
@@ -15,3 +16,39 @@ import (
...
+	}
+
+	metadata.Finalizers = append(metadata.Finalizers, finalizer)
+	return true
+}
+
+// RemoveFinalizerInMetadata removes the finalizer from the finalizers in ObjectMeta.
+// If it exists, remove the finalizer from the slice.
+// Returns true if the ObjectMeta has been changed.
+func RemoveFinalizerInMetadata(metadata *metav1.ObjectMeta, finalizer string) bool {
```
pmalek, line 37: If the newly introduced type is to have a purpose, it should be accepted in here (the same above in `EnsureFinalizerInMetadata`

```suggestion
// RemoveFinalizerInMetadata removes the finalizer from the finalizers in ObjectMeta.
// If it exists, remove the finalizer from the slice.
// Returns true if the ObjectMeta has been changed.
func RemoveFinalizerInMetadata(metadata *metav1.ObjectMeta, finalizer ControlPlaneFinalizer) bool {
```

WDYT @randmonkey ?

randmonkey: The purpose of `RemoveFinalizerInMetadata` is a general function for adding or removing finalizers, not limited to `ControlPlane`. So here I did not use `ControlPlaneFinalizer` here.

pmalek: So perhaps the type shouldn't be so narrow i.e. defined only for controlplane finalizers but generally for finalizers?

This way the funcs you've added (e.g. the one above) could use it and also in the future if there's new finalizers added for other type like e.g. `DataPlane` then it can be reused.

WDYT @randmonkey ?

randmonkey: I defined the type `ControlPlaneFinalizer` just to annotate that these are finalizers added to `ControlPlane`s by controlplane controller, but not define a type for all finalizers. If you really, really, really ... concern about the type being defined but not used, I think I could remove the definition of `ControlPlaneFinalizer`, and use string directly when call these functions in controlplane controller.

pmalek: Feel free to resolve this comment if you disagree with this approach.

I'll let others express their view on this.

shaneutt (dismissed):

shaneutt:
Looks like there's a couple small lint issues that need to be fixed before we can proceed.

randmonkey:
> Looks like there's a couple small lint issues that need to be fixed before we can proceed.

fixed. lint issues comes out since lint is enabled on test files in some recent PRs.

shaneutt (approved):

PR #149, opened 2022-08-10
fix: add flag for cluster CA secret namespace and check if it's provided

pmalek:

Up until now there was no way to specify the namespace for secret containing generated cluster CA.

This PR introduces the mechanism which first check the `POD_NAMESPACE` environmental variable (which is normally provided in k8s environment) and then falls back to this flag - `-cluster-ca-secret-namespace`. If both are empty it exits with an error since the cluster CA secret cannot be created without providing a namespace.


internal/manager/run.go
```diff
@@ -215,14 +215,20 @@ type caManager struct {
...
+	return m.maybeCreateCACertificate(ctx)
 }
 
-func (m *caManager) maybeCreateCACertificate() error {
+func (m *caManager) maybeCreateCACertificate(ctx context.Context) error {
 	// TODO https://github.com/Kong/gateway-operator/issues/108 this also needs to check if the CA is expired and
 	// managed, and needs to reissue it (and all issued certificates) if so
 	ca := &corev1.Secret{}
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	ctx, cancel := context.WithTimeout(ctx, time.Second*30)
```
shaneutt, line 231: :+1:

shaneutt (commented):
So as far as the code goes everything looks good in terms of getting the desired effect, but what I'm not clear on is the why we're enabling this. What was the impetus for this change?

rainest (approved):
> What was the impetus for this change?

Running outside a cluster, where the downward API won't work (and we haven't set it manually in tests).

shaneutt (dismissed):

shaneutt (approved):

PR #152, opened 2022-08-11
Test reference environment variables.

### LINKED ISSUE #48: Handle EnvFrom
### Problem Statement

When working on the POC there are several places where equality and sorting for types (like `DeploymentOptions`) was implemented _without_ support for `corev1.EnvFromSource`. This was done for speed because it was lot of leg work to add the functionality for this type, but now's the time to get that rectified.

### Acceptance Criteria

- [ ] As a user I can use EnvFrom for my `GatewayConfiguration`, `ControlPlane`, and `DataPlane` environment variables and my entries do not get overwritten and configuration can be updated appropriately.

### Additional Notes

- [ ] double check that we're handling ValueFrom in regular `EnvVar` correctly in places where we're using it
- [ ] TODO items which reference this issue have been cleaned up

rainest:
Add tests that confirm transfer of reference EnvVars to child resources.

Update test suites to use type:name format for clusters and support GKE.

Disable cluster cleanup for pre-existing clusters.

**Which issue this PR fixes**
Fix #48, sort of. This only adds a test similar to the existing non-reference test. AFAICT, no code needs to change.

We already use an array of the standard [EnvVar](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#envvar-v1-core) struct in GatewayConfiguration/ControlPlane/DataPlane resources, and copy the user-provided value over as-is into the Deployments we generate after. The standard EnvVar already supports either `value` or `valueFrom`, and the code to bind that to a container environment is part of the standard Kubernetes Deployment/Pod implementation. Copying that stock type as-is supports `EnvVarSource` without additional code from us.

Issue is a bit ambiguous on whether it's talking about `EnvVarSource` (the value of some individual variable is obtained via reference) or `EnvFrom` (an entire set of variables is obtained from keys in some referenced resource), but we also use and copy the stock type for the latter.

There are no TODOs referencing #48.

The not doing anything significant here has been hijacked to align the test environments with the ingress controller, namely:
- If you provide an existing cluster, you can reuse it. Cluster cleanup now only runs if the cluster was generated by the test suite. 
[test_reuse.txt](https://github.com/Kong/gateway-operator/files/9310811/test_reuse.txt)
- Existing cluster names use the `<type>:<name>` format instead of just `<name>`.

Mostly I copied the test cluster setup from the ingress controller suite_test.go, including the GKE setup, and renamed some things. Didn't actually test the GKE setup independently here.

**PR Readiness Checklist**:

- [ ] ~the `CHANGELOG.md` release notes have been updated to reflect significant changes~ no user-facing change.


test/e2e/suite_test.go
```diff
@@ -70,36 +72,72 @@ func TestMain(m *testing.M) {
...
+
+	///if imageLoad != "" {
+	///	imageLoader, err := loadimage.NewBuilder().WithImage(imageLoad)
+	///	exitOnErr(err)
+	///	fmt.Println("INFO: load image", imageLoad)
+	///	addons = append(addons, imageLoader.Build())
+	///}
+
+	///env, err = envBuilder.WithAddons(addons...).Build(ctx)
+	///exitOnErr(err)
```
shaneutt, line 103: It would appear we need to do something with this commented out code.

test/integration/gatewayconfiguration_test.go
```diff
@@ -20,8 +20,10 @@ import (
 )
 
 const (
-	testEnvVar = "KONG_INTEGRATION_TESTS"
-	testEnvVal = "TEST_VALUE"
+	testEnvVar         = "KONG_INTEGRATION_TESTS"
+	testEnvVal         = "TEST_VALUE"
+	testEnvVarFromName = "KONG_INTEGRATION_TESTS_FROM"
+	testEnvVarFromKV   = "dzhambul"
```
shaneutt, line 26: Does this mean something specific?

test/e2e/suite_test.go
rainest: Committing without having actually saved a change, my dearest friend.

test/integration/gatewayconfiguration_test.go
rainest: It's mostly just to have some value that we can guarantee is the same in the check and in ConfigMap creation, same as the other. Arbitrarily, it is the former name of a city in Kazakhstan, because I was looking at an old map when I wanted to find a test value.

shaneutt (approved):

PR #155, opened 2022-08-12
chore: don't send telemetry from dev builds

pmalek:

Prevents sending telemetry data from dev builds.


main.go
```diff
@@ -91,9 +90,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
```
pmalek, line 94: This is already done in `internal/manager/run.go`

mlavacca (approved):

Issue #34, opened 2022-06-07
Auto-disable `Gateway` support for clusters missing the APIs


shaneutt:
### Problem Statement

Currently the CRDs for [Gateway API](https://github.com/kubernetes-sigs/gateway-api) must be applied to a cluster before deploying the Operator or else errors will ensue. Preferably the operator would be able to start and still work on `ControlPlanes` and `DataPlanes` even if those CRDs aren't present, and would provide information about the problem and try to auto-correct for it over time.

### Acceptance Criteria

- [ ] a flag is available to deploy the Gateway Operator without Gateway API support, when used the operator doesn't crash when the Gateway API CRDs are missing and enables raw use of `DataPlanes` and `ControlPlanes` in the absence of `Gateways` (suggestion `--with-gateway-controller-disabled`)

PR #156, opened 2022-08-12
feat: require Gateway CRD for controller

rainest:
Add flags to manually disable controllers.

Disable Gateway controller if CRD not present.

**Which issue this PR fixes**
#34 

~This copies the similar controller code almost verbatim aside from the `setupControllers()` list of specific controllers. The code is fairly brief and isn't likely to change much, so we're probably fine if we do just copy/paste here. Do we want some sort of shared library?~ This is small enough that we're just going to leave it here pending some shared library to move it into. In this particular case, we'd like to see if we can upstream it: https://github.com/Kong/team-k8s/issues/240

**PR Readiness Checklist**:


internal/manager/run.go
```diff
@@ -143,26 +146,17 @@ func Run(cfg Config) error {
...
-	if err = (&controllers.ControlPlaneReconciler{
-		Client:          mgr.GetClient(),
-		Scheme:          mgr.GetScheme(),
-		ClusterCASecret: cfg.ClusterCASecretName,
-	}).SetupWithManager(mgr); err != nil {
-		return fmt.Errorf("unable to create controller ControlPlane: %w", err)
+	// load controllers
+	controllers := setupControllers(mgr, &cfg)
+	if err != nil {
+		return fmt.Errorf("unable to setup controller as expected %w", err)
```
pmalek, line 152: It seems that this isn't 100% correct, i.e. `setupControllers` doesn't return an error and this `if` statement here checks the previously assigned error: from `mgr.Add()`

rainest: Yeah, good catch, that's a leftover from the KIC variant, which can return an error from its version of this function.

internal/manager/controller_setup.go
```diff
@@ -0,0 +1,110 @@
...
+// Controller Manager - Controller Definition Interfaces
+// -----------------------------------------------------------------------------
+
+// Controller is a Kubernetes controller that can be plugged into Manager.
+type Controller interface {
+	SetupWithManager(ctrl.Manager) error
+}
+
+// AutoHandler decides whether the specific controller shall be enabled (true) or disabled (false).
+type AutoHandler func(client.Client) bool
```
pmalek, line 25: I think - especially that this is exported - that this might use a slightly different name, similar to the one used below: `CRDChecker` :)

And even better if that was an interface, then the code below would still work but we're open for other implementations not limited to `func`s.

pmalek (commented):
I like this in general. Left a nit as food for thought.

internal/manager/controller_setup.go
rainest: Added the interface point to the new issue to try and port this upstream. Would prefer to leave this as-is here though, since I'd like it to be the same as the code it was copied from despite the bad name.

test/integration/suite_test.go
```diff
@@ -246,6 +246,9 @@ func startControllerManager() {
 	cfg.LeaderElection = false
 	cfg.DevelopmentMode = true
 	cfg.ControllerName = "konghq.com/gateway-operator-integration-tests"
+	cfg.GatewayControllerEnabled = true
+	cfg.ControlPlaneControllerEnabled = true
+	cfg.DataPlaneControllerEnabled = true
```
pmalek, line 251: Can you put those into the `DefaultConfig`?

Sidenote: I have a feeling that we should make the default config a return value from a `func` instead of having it defined as is.

rainest: :+1:

internal/manager/run.go
```diff
@@ -90,18 +89,27 @@ type Config struct {
...
-	MetricsAddr:         ":8080",
-	ProbeAddr:           ":8081",
-	WebhookPort:         9443,
-	DevelopmentMode:     false,
-	LeaderElection:      true,
-	ClusterCASecretName: "kong-operator-ca",
-	// TODO: Extract this into a named const and use it in all the placed where
-	// "kong-system" is used verbatim: https://github.com/Kong/gateway-operator/pull/149.
-	ClusterCASecretNamespace: "kong-system",
+func DefaultConfig() Config {
```
pmalek, line 98: Oh cool 👍

pmalek (approved):

PR #157, opened 2022-08-15
docs: update KEP 1

shaneutt:

At this point nobody on the team has had any significant opposition to us moving from POC to our first alpha release with the Gateway Operator according to the plan so far laid out in the design and graduation criteria of the KEP. So this PR is just some small housekeeping to mark the KEP as implementable and change some milestone language.


pmalek (approved):
🚀

PR #160, opened 2022-08-16
ci: make dev build flag a build variable

pmalek:

This PR makes the dev build flag a build flag so that we can control at the build stage whether a build will result in a dev or release build.

The behavior of [being able to control this with an env variable] is retained(https://github.com/Kong/gateway-operator/blob/ad707aaa6fe4dac634ea4eac41fe0fad9242937b/main.go#L57-L60).


pmalek:
Related #161

pmalek:
At this moment we've decided to not pursue the path of adding yet another build path but to accept the possible noise that we might encounter in telemetry data (given that users can run the operator without the development flag/env var - `CONTROLLER_DEVELOPMENT_MODE=true` - set)

PR #164, opened 2022-08-18
fix: bind flags once and use the bound logger flags for logger configuration

pmalek:

This PR:

* fixes flags parsing - parses them just once instead of twice
* makes the parsed logger flags actually be used for logger configuration.
* adds `-cluster-ca-secret-namespace kong-system` flag for local debugging (run via `make run` and `make debug`) which is needed now to run outside a cluster

So now users will get those (hidden until now) flags at their disposal:

```
  -zap-devel
        Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)
  -zap-encoder value
        Zap log encoding (one of 'json' or 'console')
  -zap-log-level value
        Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', or any integer value > 0 which corresponds to custom debug levels of increasing verbosity
  -zap-stacktrace-level value
        Zap Level at and above which stacktraces are captured (one of 'info', 'error', 'panic').
  -zap-time-encoding value
        Zap time encoding (one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano'). Defaults to 'epoch'.

```


Makefile
```diff
@@ -279,11 +279,15 @@ endif
...
+	CONTROLLER_DEVELOPMENT_MODE=true go run ./main.go --no-leader-election \
+		-cluster-ca-secret-namespace kong-system \
+		-zap-time-encoding iso8601
 
 .PHONY: debug
 debug: manifests generate fmt vet install
-	CONTROLLER_DEVELOPMENT_MODE=true dlv debug ./main.go -- --no-leader-election
+	CONTROLLER_DEVELOPMENT_MODE=true dlv debug ./main.go -- --no-leader-election \
+		-cluster-ca-secret-namespace kong-system \
+		-zap-time-encoding iso8601
```
pmalek, line 290: Add a sane default for local debugging. Feel free to comment.

shaneutt (dismissed):

mlavacca (approved):

Issue #24, opened 2022-06-07
Handle Hostnames for Gateway Status


shaneutt:
### Problem Statement

The original POC only bothered dealing with IPs for `Gateway` status since that's all that was available in a `ktf` testing environment. We need to make sure that we have proper support for Hostnames which will be provided by some providers such as AWS EKS.

### Prerequisites

- [ ] https://github.com/Kong/kubernetes-testing-framework/issues/285

### Acceptance Criteria

- [ ] when a `LoadBalancer` service provides hosts but not IPs, this is supported and reflected in `Gateway` resource statuses
- [ ] an E2E test exists which verifies host functionality against AWS EKS clusters `LoadBalancer` `Service` implementation

### Additional Notes

- [ ] TODO comments which linked to this issue are removed once resolved

mflendrich:
@shaneutt what is the blocker here?

mflendrich:
After a more careful look: is EKS really a prerequisite for this?

shaneutt:
> After a more careful look: is EKS really a prerequisite for this?

It's a simple way to test host based LB provisioning, while also having the valuable side effect of getting us testing on AWS which is one of the most prominent cloud providers. I'm certain we can find another way to do it, but given the benefits it seemed to me that this way was the most value for time spent.

rainest:
Having the additional test environment available on KTF is good in general, but we can refactor the functions here to allow unit testing the actual status translation.

LoadBalancerStatus is standard API type and the [individual entries](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#loadbalanceringress-v1-core) are quite simple: for our purposes, they're just a collection of zero, one, or two string fields that we copy elsewhere as-is: we can easily cover all possible combinations and don't need to care about any weird idiosyncrasies in the values providers actually use, just whether the field is or is not present.

PR #175, opened 2022-08-18
feat: expand Gateway status support

rainest:

Add all addresses in the Gateway Service LoadBalancer status to the Gateway address status, not only the first.

Support hostname Gatway status addresses. Address conversion is now split into a separate function to let you unit test it.

Only add a ClusterIP address if the Gateway Service is not a LoadBalancer type Service. This aligns with our current unmanaged handling in the ingress controller. If you have indicated that you want a LoadBalancer, and use this status information to inform DNS or whatever, you probably do want that to reflect an externally-accessible address, and having an internal address temporarily shown won't like be of much use (and could actually cause issues, since it may be stuck in DNS until TTL expires). 

**Which issue this PR fixes**
Satisfies the first AC (support hostname statsuses) of #24, but not the AWS test AC.

- Service status entries with both IP and hostname values are valid (insofar as you can make one with neither and the API server acceps it), so they are supported by adding _two_ entries to Gateway status (which can only have a single type per entry).
- Service status entries with neither an IP nor hostname value are similarly valid but cannot meaningfully translate to a Gateway status address. I just silently skip these, since I figure they're the only thing you have (so you have "zero" addresses and the Gateway does not become ready) or you have a mix of the two, in which case setting only the hostname/IP entries in the Gateway status is... probably okay?
- `ensureGatewayMarkedReady()` originally issued its own `r.Client.Status().Update()` call if the Gateway was ready, but it looks like this was unnecessary as the reconciler always calls `updateStatus()` after. I've removed the call within `ensureGatewayMarkedReady()`
- `ensureGatewayMarkedReady()` now clears out existing status entries if encounters any failure other than an inability to list Services (that doesn't necessarily indicate a Service status change--it's most likely just a transient issue talking to the API server). Originally it looked like we could enter wonky state transitions where you'd be marked not ready, but still have leftover addresses from a previous ready state.

**PR Readiness Checklist**:

- [ ] ~the `CHANGELOG.md` release notes have been updated to reflect significant changes~ alpha release yeaaaaaaaaaaah!


controllers/gateway_controller_reconciler_utils.go
```diff
@@ -87,42 +87,58 @@ func (r *GatewayReconciler) ensureGatewayMarkedReady(ctx context.Context, gatewa
 
 		count := len(services)
 		if count > 1 {
+			gateway.Status.Addresses = []gatewayv1alpha2.GatewayAddress{}
 			return fmt.Errorf("found %d services for DataPlane currently unsupported: expected 1 or less", count)
```
pmalek, line 91: I do reckon you're not changing this line but while we're here, how about:

```suggestion
			return fmt.Errorf("found %d services for DataPlane %s/%s, currently unsupported: expected 1 or less",
				count, dataplane.Namespace, dataplane.Name,
			)
```

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -87,42 +87,58 @@ func (r *GatewayReconciler) ensureGatewayMarkedReady(ctx context.Context, gatewa
 
 		count := len(services)
 		if count > 1 {
+			gateway.Status.Addresses = []gatewayv1alpha2.GatewayAddress{}
```
pmalek, line 90: Can we add a comment on why we set this? Similarly below

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -87,42 +87,58 @@ func (r *GatewayReconciler) ensureGatewayMarkedReady(ctx context.Context, gatewa
...
+	}
+	return nil
+}
 
-		gatewayIPs := make([]string, 0)
-		if len(svc.Status.LoadBalancer.Ingress) > 0 {
-			ip := svc.Status.LoadBalancer.Ingress[0].IP
-			if ip == "" {
-				return errors.New("missing loadbalancer.ingress[0].ip")
+func gatewayAddressesFromService(svc corev1.Service) ([]gatewayv1alpha2.GatewayAddress, error) {
```
pmalek, line 111: Can we add a comment on why LoadBalancer addresses are added but for other service types we're only adding cluster IPs?

rainest:
Reviewing this further I think it could use another pass to:
- Disentangle responsibilities between the reconciler and helpers. We currently have some situations where the helper sets the ready condition and some where the reconciler does it.
- Rename functions whose name includes "Ready". "Ready" can mean either "the state of being ready" or, here, "a Condition with type Ready", which may indicate that a resource's state _is not_ ready, so we have functions whose names suggest the former even though they also handle not ready states (e.g. `SetReady`), and some that actually are the former (e.g. `IsReady`).
- Update readiness and status information always. Although this PR does start clearing Gateway addresses when the Gateway is not ready, it doing so is contingent on _something else_ setting a false Condition to trigger an update. Some of the states reviewed by `ensureGatewayMarkedReady()` should trigger a not ready state independently.
- Review the readiness criteria. It looks like we just require no false Conditions, but for some Conditions false is the good version, e.g. ListenerConditionConflicted.

shaneutt:
@rainest looks like you need to sign the updated CLA.

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -85,44 +85,68 @@ func (r *GatewayReconciler) ensureGatewayMarkedReady(ctx context.Context, gatewa
...
-
-		newAddresses := make([]gatewayv1alpha2.GatewayAddress, 0, len(gatewayIPs))
-		ipaddrT := gatewayv1alpha2.IPAddressType
-		for _, ip := range append(gatewayIPs, svc.Spec.ClusterIP) {
-			newAddresses = append(newAddresses, gatewayv1alpha2.GatewayAddress{
-				Type:  &ipaddrT,
-				Value: ip,
-			})
+	} else {
+		// if the Service is not a LoadBalancer, it will never have any public addresses and its status address list
```
shaneutt, line 138: Technically we could allow for a `NodePort` type here. We don't really have the spec to enable it at the moment, so we don't need to stop and add it but I think a TODO and an issue for allowing it would be helpful as it's inevitable someone is going to want it at some point.

shaneutt (changes requested):
Seeing a couple issues, including updating status before returning an error. In general, even if we make that "work" I believe we should avoid it because it's too subtle. Better to handle the status at the caller level in a concise error checking switch.

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -85,44 +85,68 @@ func (r *GatewayReconciler) ensureGatewayMarkedReady(ctx context.Context, gatewa
...
 		}
 
+		// once we've successfully retrieved the proxy Service, any issue we detect indicates that the Service is not
+		// currently ready to handle traffic, so we clear out any existing addresses that were set based on an earlier
+		// Service state
 		count := len(services)
 		if count > 1 {
-			return fmt.Errorf("found %d services for DataPlane currently unsupported: expected 1 or less", count)
+			gateway.Status.Addresses = []gatewayv1alpha2.GatewayAddress{}
+			return fmt.Errorf("found %d services for DataPlane %s/%s currently unsupported: expected 1 or less", count,
```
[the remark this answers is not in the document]
rainest: We are handling status _updates_ in the caller already unless I'm missing something--this is the third bullet in the special notes section. Our error handling for `ensureGatewayMarkedReady()` does not abort, it sets conditions (aka more status). We always call `updateStatus()` after, and it [performs an update if needed](https://github.com/Kong/gateway-operator/blob/b78a36650b272698019bdb5e08b044b379da232c/controllers/gateway_controller.go#L306-L317). This is basically demonstrative of the whole "these things have poor unclear separation of concerns" issue in #192. Since the existing condition sets in `Reconcile()` would have required another update anyway, I went with using the `updateStatus()` call only.

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -85,44 +85,68 @@ func (r *GatewayReconciler) ensureGatewayMarkedReady(ctx context.Context, gatewa
...
+}
 
-		gatewayIPs := make([]string, 0)
-		if len(svc.Status.LoadBalancer.Ingress) > 0 {
-			ip := svc.Status.LoadBalancer.Ingress[0].IP
-			if ip == "" {
-				return errors.New("missing loadbalancer.ingress[0].ip")
+func gatewayAddressesFromService(svc corev1.Service) ([]gatewayv1alpha2.GatewayAddress, error) {
+	IPAddressType := gatewayv1alpha2.IPAddressType
+	HostnameAddressType := gatewayv1alpha2.HostnameAddressType
```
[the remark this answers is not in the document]
rainest: ??? the whole reason we have these is because we can't get the address of the constants upstream, or any constants, for that matter--duplicating them in our packages wouldn't make it any better.

controllers/gateway_controller_reconciler_utils.go
rainest: Can we? NodePorts don't themselves have public addresses--the Pod addresses provide some level of access, but we don't know that they're actually the addresses clients will want like we do with LoadBalancers. The approach we should take for anything other than LoadBalancer is to basically duplicate the KIC `--publish-status-address` feature where we stuff whatever you give us (e.g. the address of some externally-defined LB that distributes traffic among the node addresses) into the status fields and call it a day.

rainest:
https://github.com/Kong/gateway-operator/pull/226 merging basically obliterated everything this was going to modify, so back to draft it go.

controllers/gateway_controller_reconciler_utils.go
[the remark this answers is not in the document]
shaneutt: I mis-typed, I meant globals!

controllers/gateway_controller_reconciler_utils.go
shaneutt: A spike to remind ourselves to check on it: https://github.com/Kong/gateway-operator/issues/260

controllers/gateway_controller_reconciler_utils.go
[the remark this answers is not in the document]
shaneutt: As per my comment above, if we're doing that I think it's too subtle and should change. Doesn't need to be done in this PR, but I would personally like to see a follow up for it so we take another look at it later during some maintenance.

PR #180, opened 2022-08-19
fix: predicates for detecting whether the deployment is active

mlavacca:

In our tests, we use a pair of predicates that check whether the controlplane or dataplane deployments are ready. To be sure about that, we need to check that the number of active replicas is equal to the number of replicas.

**Which issue this PR fixes**

**PR Readiness Checklist**:


pmalek (approved):
Out of curiosity: did this cause you problems somehow? You had more available replicas than ready replicas somehow?

mlavacca:
> Out of curiosity: did this cause you problems somehow? You had more available replicas than ready replicas somehow?

Yep, I experienced a state in which there was 1 replica of the controlplane (the deployment was just scaled up by the controller), but 0 ready replicas. In that situation the conditional we had before returned true even if no pod was ready yet.

Issue #161, opened 2022-08-16
Inject version metadata (via `ldflags`) in Dockerfile


pmalek:
~~After #160 gets merged we need to manage the DEV_BUILD flag in our Dockerfiles.~~

This issue tracks the effort to use makefile targets (or do this manually via `go build -ldflags ...`) to inject version info variable [in Dockerfile during build](https://github.com/Kong/gateway-operator/blob/1fbc71315946aad73b3f083a10a01f8913df3c00/Dockerfile#L29) since as of now, no version information is embedded in the binary.

pmalek:
Fixed by #195

PR #195, opened 2022-08-23
ci: add version info to container image

pmalek:

This PR adds build info to operator's container image.

* It adds build args to `Dockerfile`
* It makes it so that build step in `Dockerfile` uses a build instruction from `Makefile`
* It adds a `-v` flag which will print the version information embedded in `metadata` package

**Which issue this PR fixes**

#161 

Do we want to mount `.git` and `Makefile` and rely on the tag/version information retrieved from `git` command invocation or do we want to inject it somehow?

This is not a problem when using GH Actions as defined in https://github.com/Kong/gateway-operator/blob/3af4fb1943cf10545672db8f547eb574d157ecc6/.github/workflows/release.yaml but if devs want to test container image build process then we'd need to come up with a solution.

One of those (solutions) is in this PR, namely adding `.git` and `Makefile` mounts and relying on git information that is injected in `build.operator.image` target.


shaneutt (approved):
I'm fine with what we're adding here, as for your notes about dev builds: I imagine we can deal with that once it starts bothering someone.

Issue #197, opened 2022-08-23
Flaky test: `TestGatewayEssentials`


pmalek:
The following failure has been observed in this CI run: https://github.com/Kong/gateway-operator/runs/7970955157?check_suite_focus=true#step:5:3743

```    gateway_test.go:99: 
        	Error Trace:	/home/runner/work/gateway-operator/gateway-operator/test/integration/gateway_test.go:99
        	Error:      	Condition never satisfied
        	Test:       	TestGatewayEssentials
```

The test succeeded on a subsequent run without any code modifications: https://github.com/Kong/gateway-operator/runs/7971300586?check_suite_focus=true

pmalek:
Second failure in https://github.com/Kong/gateway-operator/runs/7971753639?check_suite_focus=true#step:5:4631

shaneutt:
Saw it again here: https://pipelines.actions.githubusercontent.com/serviceHosts/c81d839e-81ff-4819-b743-a1ded9d25776/_apis/pipelines/1/runs/2022/signedlogcontent/2?urlExpires=2022-08-23T15%3A07%3A15.9979174Z&urlSigningMethod=HMACV1&urlSignature=dMOfZTomyII2r7JIU4YCYXCGryE3qJUzd9BuTsvAY84%3D

After debugging, I'm pretty sure #91 introduced the flake because the 10 seconds we used to wait is not enough now that we're deleting and replacing the `DataPlane` just before. My suspicion is that a slight rework of the test to allow more time during these steps will be adequate, I'll create a patch.

pmalek:
> Saw it again here: https://pipelines.actions.githubusercontent.com/serviceHosts/c81d839e-81ff-4819-b743-a1ded9d25776/_apis/pipelines/1/runs/2022/signedlogcontent/2?urlExpires=2022-08-23T15%3A07%3A15.9979174Z&urlSigningMethod=HMACV1&urlSignature=dMOfZTomyII2r7JIU4YCYXCGryE3qJUzd9BuTsvAY84%3D

Weird URL 🤔 And it says `{"count":11,"value":"Uri expired"}`

> After debugging, I'm pretty sure https://github.com/Kong/gateway-operator/pull/91 introduced the flake because we now delete and try to have the DataPlane re-created during TestGatewayEssentials. I'm going to verify whether the DataPlane is simply not given enough time, OR if it's actually perma-crashed.

Oh, nice. Thanks!

shaneutt:
[...]

PR #207, opened 2022-08-23
fix: extend subresource readiness wait in tests

shaneutt:

While investigating https://github.com/Kong/gateway-operator/issues/197 I noticed that we have a sub-resource readiness wait in the tests that provides only the exact same amount of time as we provide a timeout for our default http client: 10 seconds. This means when connections hang, retries will likely not occur which is not an intended effect. This patch simply triples the wait time to 30s, as this gives time for at least 3 retries before termination.

This doesn't necessarily resolve https://github.com/Kong/gateway-operator/issues/197, it may still be the case that the replacement of `DataPlanes` is not yet stable. I'll keep investigating that separately.


pmalek (approved):
🚀

PR #218, opened 2022-08-23
feat: keep Deployment environments updated

### LINKED ISSUE #97: Update dataplane env vars in controlplane when dataplane service changes

### LINKED ISSUE #196: Data race on setting the Cloud Flare logger singleton from multiple tests

### Current Behavior

When running `make test.integration` I encountered the followig:

```
==================
WARNING: DATA RACE
Write at 0x000108380f20 by goroutine 703:
  github.com/cloudflare/cfssl/log.SetLogger()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/github.com/cloudflare/cfssl@v1.6.1/log/log.go:62 +0x45c
  github.com/kong/gateway-operator/controllers.signCertificate()
      /Users/USER/code_/gateway-operator/controllers/utils.go:110 +0x28
  github.com/kong/gateway-operator/controllers.maybeCreateCertificateSecret()
      /Users/USER/code_/gateway-operator/controllers/utils.go:219 +0x9c0
  github.com/kong/gateway-operator/controllers.(*DataPlaneReconciler).ensureCertificate()
      /Users/USER/code_/gateway-operator/controllers/dataplane_controller_reconciler_utils.go:114 +0x190
  github.com/kong/gateway-operator/controllers.(*DataPlaneReconciler).Reconcile()
      /Users/USER/code_/gateway-operator/controllers/dataplane_controller.go:120 +0x578
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:121 +0xe4
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:320 +0x360
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:273 +0x240
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:234 +0x98

Previous write at 0x000108380f20 by goroutine 704:
  github.com/cloudflare/cfssl/log.SetLogger()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/github.com/cloudflare/cfssl@v1.6.1/log/log.go:62 +0x45c
  github.com/kong/gateway-operator/controllers.signCertificate()
      /Users/USER/code_/gateway-operator/controllers/utils.go:110 +0x28
  github.com/kong/gateway-operator/controllers.maybeCreateCertificateSecret()
      /Users/USER/code_/gateway-operator/controllers/utils.go:219 +0x9c0
  github.com/kong/gateway-operator/controllers.(*ControlPlaneReconciler).ensureCertificate()
      /Users/USER/code_/gateway-operator/controllers/controlplane_controller_reconciler_utils.go:290 +0x19c
  github.com/kong/gateway-operator/controllers.(*ControlPlaneReconciler).Reconcile()
      /Users/USER/code_/gateway-operator/controllers/controlplane_controller.go:234 +0x1330
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:121 +0xe4
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:320 +0x360
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:273 +0x240
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:234 +0x98

Goroutine 703 (running) created at:
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:230 +0x364
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:241 +0x254
  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func1()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/manager/runnable_group.go:219 +0x158
  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func2()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/manager/runnable_group.go:222 +0x44

Goroutine 704 (running) created at:
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:230 +0x364
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:241 +0x254
  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func1()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/manager/runnable_group.go:219 +0x158
  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func2()
      /Users/USER/.gvm/pkgsets/go1.19/global/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/manager/runnable_group.go:222 +0x44
==================
```

which points to the following line in code: https://github.com/Kong/gateway-operator/blob/efea71c1b334f4eb5d9be0e845127f54a2600834/controllers/utils.go#L110 which sets a singleton to a designated value in here: https://github.com/cloudflare/cfssl/blob/7614d6cad35dd6d33c8c1fc2e2db5d9ce111e56b/log/log.go#L62

### Expected Behavior

No data race happens.

### Steps To Reproduce

```markdown
1. Run `make test.integration`
2. Wait for data race to occur
```

### Kong Ingress Controller version

```shell
N/A
```

### Kubernetes version

```shell
N/A
```

### Anything else?

_No response_

rainest:
Update the Deployment associated with a ControlPlane or DataPlane if its current actual EnvVar set differs from the set generated by the reconciler. #97 doesn't strictly need the DataPlane to do this, but the two should be consistent and we [want this UX anyway](https://github.com/Kong/gateway-operator/issues/97#issuecomment-1224155582).

Add Service information to DataPlane status.

Have the ControlPlane reconciler watch for changes to the controlled DataPlane. If the DataPlane status contains a new Service, the generated EnvVars will differ and the reconciler will update the ControlPlane with the new publish service.

Fix an unrelated race condition that broke tests.

**Which issue this PR fixes**
Fix #97 
Fix #196 

**PR Readiness Checklist**:

- [ ] ~the `CHANGELOG.md` release notes have been updated to reflect significant changes~ alphazone


apis/v1alpha1/dataplane_types.go
```diff
@@ -66,6 +66,9 @@ type DataPlaneStatus struct {
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=8
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// Service indicates the Service that exposes the DataPlane's configured routes
+	Service string `json:"service,omitempty"`
```
shaneutt, line 71: I'm not necessarily opposed to this, but I would like to talk through this some.

One of the implications of adding this `Service` reference would seem to be that the `DataPlane` has some kind of "ownership" over the `Service`, which it does not: ownership is maintained by the `Gateway`. In the current state where the `Gateway` owns the `Service`, does this field have any meaning in the context of directly using the `DataPlane` API?

What I'm starting to realize is that the `Service` should actually belong to the `DataPlane`, especially if we add a reference like this. Are you opposed to moving the `Service` logic from the `Gateway` controller to the `DataPlane` controller so that this works when a caller directly uses the `DataPlane` API? Can you see any downsides of doing so?

rainest: Wait, what? I don't follow, that doesn't seem to align with the current code. The DataPlane controller does [create its Service](https://github.com/Kong/gateway-operator/blob/758facd52c2e5eb1f5b1a4f37be621376f422eb6/controllers/dataplane_controller.go#L79-L80), and the standalone DataPlane test [checks for its existence](https://github.com/Kong/gateway-operator/blob/758facd52c2e5eb1f5b1a4f37be621376f422eb6/test/integration/dataplane_test.go#L69-L71).

The Gateway controller's interaction with it is just to retrieve the Service associated with a DataPlane and then provide it to the ControlPlane builder and its own status setter.

shaneutt: Whoops! :facepalm: my bad sorry.

controllers/controlplane_controller.go
```diff
@@ -156,13 +160,23 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil // no need to requeue, status update will requeue
 	}
 
-	debug(log, "retrieving dataplane service info", controlplane)
-	dataplaneServiceName, err := gatewayutils.GetDataplaneServiceNameForControlplane(ctx, r.Client, controlplane)
+	debug(log, "retrieving connected dataplane", controlplane)
+	dataplane, err := gatewayutils.GetDataPlaneForControlPlane(ctx, r.Client, controlplane)
+	var dataplaneServiceName string
```
randmonkey, line 165: where is this variable `dataplaneServiceName` used except for the following `else` block? If nowhere else used, the variable should not be defined in the block.

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -166,6 +167,15 @@ func (r *ControlPlaneReconciler) ensureDeploymentForControlPlane(
 			updated = true
 		}
 
+		// We do not want to permit direct edits of the Deployment environment. Any user-supplied values should be set
+		// in the ControlPlane. If the actual Deployment environment does not match the generated environment, either
+		// something requires an update (e.g. the associated DataPlane Service changed and value generation changed the
+		// publish service configuration) or there was a manual edit we want to purge.
+		if !reflect.DeepEqual(existingDeployment.Spec.Template.Spec.Containers[0].Env, controlplane.Spec.Env) {
+			existingDeployment.Spec.Template.Spec.Containers[0].Env = controlplane.Spec.Env
```
randmonkey, line 175: We need also check `EnvFrom` here.

controllers/dataplane_controller_reconciler_utils.go
```diff
@@ -150,6 +160,15 @@ func (r *DataPlaneReconciler) ensureDeploymentForDataPlane(
 		var updated bool
 		existingDeployment := &deployments[0]
 		updated, existingDeployment.ObjectMeta = k8sutils.EnsureObjectMetaIsUpdated(existingDeployment.ObjectMeta, generatedDeployment.ObjectMeta)
+
+		// We do not want to permit direct edits of the Deployment environment. Any user-supplied values should be set
+		// in the DataPlane. If the actual Deployment environment does not match the generated environment, either
+		// something requires an update or there was a manual edit we want to purge.
+		if !reflect.DeepEqual(existingDeployment.Spec.Template.Spec.Containers[0].Env, dataplane.Spec.Env) {
```
randmonkey, line 167: We need also check `EnvFrom` here.

controllers/utils.go
```diff
@@ -53,6 +54,14 @@ func (l loggerShim) Err(msg string)     { l.logger.V(logging.InfoLevel).Info(msg
 func (l loggerShim) Crit(msg string)    { l.logger.V(logging.InfoLevel).Info(msg) }
 func (l loggerShim) Emerg(msg string)   { l.logger.V(logging.InfoLevel).Info(msg) }
 
+var caLoggerInit sync.Once
+
+func setCALogger(logger logr.Logger) {
```
randmonkey, line 59: Seems that we need to add such functions and package level `sync.Once` variables for each place where we want to add some fields inside functions, in order to avoid data race? IHAC that there would be too many such things. But for now, I think it is acceptable before we have got a better solution to loggings.

randmonkey (changes requested):

controllers/dataplane_controller_reconciler_utils.go
pmalek: BTW I do understand that we have just 1 container (kong gateway) in data plane pod but I'd be careful with things like `.Spec.Template.Spec.Containers[0]`.

Why not be on the safe side and do the traversal over the containers slice and pick the one with the gateway? My main point is that no one will remember to make this change when at some point we introduce e.g. a sidecar for whatever purposes.

rainest: Is there a convenience function for this already? I can write one but it seems like getting a container by name is something client-go would provided, I just don't know what it is if so.

controllers/utils.go
rainest: This is a special case where cfssl wants to handle its own logging instead of returning errors, so we have to stuff our own logger in. Ideally it wouldn't do that, but it's not so annoying that I'd rather write our own CA tool.

The typical case where we just call `logger := log.FromContext(ctx).WithName("MTLSCertificateCreation")` or whatever and use that doesn't need this because we're the ones handling the logs, and we use function-scoped instead of package-scoped loggers. Most libraries we use return errors that we then log, instead of using their own logging. deck is the other main exception, but that's our own damn fault.

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -16,6 +17,7 @@ import (
 	operatorv1alpha1 "github.com/kong/gateway-operator/apis/v1alpha1"
 	"github.com/kong/gateway-operator/internal/consts"
 	k8sutils "github.com/kong/gateway-operator/internal/utils/kubernetes"
+	rutils "github.com/kong/gateway-operator/internal/utils/kubernetes/resources"
```
randmonkey, line 20: There is no need to import `github.com/kong/gateway-operator/internal/utils/kubernetes/resources` as `rutils` again, it is already imported. (linter did not check this?)

controllers/controlplane_controller_reconciler_utils.go
```diff
@@ -145,6 +147,11 @@ func (r *ControlPlaneReconciler) ensureDeploymentForControlPlane(
 	if count == 1 {
 		var updated bool
 		existingDeployment := &deployments[0]
+		container := rutils.GetPodContainerByName(&existingDeployment.Spec.Template.Spec, consts.ControlPlaneControllerContainerName)
```
randmonkey, line 150: IMO it is better to return `true` and update deployment here. If we return an error here, the reconciling will get a constant failure if the deployment is modified that it did not have such a container.

controllers/dataplane_controller.go
```diff
@@ -82,7 +82,8 @@ func (r *DataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, err
 	}
 	if createdOrUpdated {
-		return ctrl.Result{}, nil // requeue will be triggered by the creation or update of the owned object
+		err = r.ensureDataPlaneServiceStatus(ctx, dataplane, dataplaneService.Name)
+		return ctrl.Result{}, err // requeue will be triggered by the creation or update of the owned object
```
randmonkey, line 86: Is `return ctrl.Result{}, r.ensureDataPlaneServiceStatus(ctx, dataplane, dataplaneService.Name)` better?

test/integration/controlplane_test.go
```diff
@@ -19,6 +19,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	operatorv1alpha1 "github.com/kong/gateway-operator/apis/v1alpha1"
+	"github.com/kong/gateway-operator/internal/consts"
+	rutils "github.com/kong/gateway-operator/internal/utils/kubernetes/resources"
```
randmonkey, line 23: for consistency, Is it OK to import `github.com/kong/gateway-operator/internal/utils/kubernetes/resources` as `k8sresources` here?

test/integration/gateway_test.go
```diff
@@ -19,6 +20,7 @@ import (
 	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 
 	"github.com/kong/gateway-operator/internal/consts"
+	rutils "github.com/kong/gateway-operator/internal/utils/kubernetes/resources"
```
randmonkey, line 23: ditto here.

controllers/dataplane_controller_reconciler_utils.go
```diff
@@ -12,6 +13,7 @@ import (
 	operatorv1alpha1 "github.com/kong/gateway-operator/apis/v1alpha1"
 	"github.com/kong/gateway-operator/internal/consts"
 	k8sutils "github.com/kong/gateway-operator/internal/utils/kubernetes"
+	rutils "github.com/kong/gateway-operator/internal/utils/kubernetes/resources"
```
randmonkey, line 16: For consistency, I think it is better to import as `k8sresources` here,

controllers/controlplane_controller_reconciler_utils.go
rainest: Yeah, I missed the existing import/rename convention, thanks.

controllers/controlplane_controller_reconciler_utils.go
rainest: Is there a reason to actually return here? I think we'd only want that if there was some reason we couldn't proceed with the other checks/updates after.

I think we want what https://github.com/Kong/gateway-operator/pull/218/commits/6d40d578e99a6329f6d913af4d1bf10780ab872c adds, where it's just the same as the other things that set `updated=true`. We have to rely on the generated containers always having that container, but if it doesn't we've either done something wrong or so weird that we need to refactor this section anyway.

We could kinda skip that requirement if we update immediately here and return, but not really since if we're generating Deployments without that container it'll just get stuck in a loop running that update repeatedly

randmonkey (approved):

PR #225, opened 2022-08-24
chore: finalizers added through a single update

mlavacca:

The controlplane controller needs to attach two distinct finalizers to the controlplane. They are attached subsequently, therefore there is no need to perform two different object updates (and two reconciliations) when both of them can be applied through a single update operation.

**Which issue this PR fixes**

**PR Readiness Checklist**:


pmalek (approved):
nice 🚢

pmalek (approved):

Issue #203, opened 2022-08-23
persistent `the object has been modified; please apply your changes to the latest version and try again`


mflendrich:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Having followed the `README.md` installation instructions and applied the below yaml, I'm getting the reconciliation error from the subject repeatedly.

```yaml
kind: GatewayClass
apiVersion: gateway.networking.k8s.io/v1alpha2
metadata:
  name: kong
spec:
  controllerName: konghq.com/gateway-operator
---
kind: Gateway
apiVersion: gateway.networking.k8s.io/v1alpha2
metadata:
  name: kong
spec:
  gatewayClassName: kong
  listeners:
  - name: http
    protocol: HTTP
    port: 80
```

```console
michal@piec ~/r/K/gateway-operator (main)> kc logs -n kong-system gateway-operator-controller-manager-ddc879b87-xcrcn -f
{"level":"info","ts":1661258810.5017073,"logger":"controller-runtime.metrics","msg":"Metrics server is starting to listen","addr":"127.0.0.1:8080"}
{"level":"info","ts":1661258810.502056,"logger":"controller-runtime.webhook","msg":"Registering webhook","path":"/validate"}
{"level":"info","ts":1661258810.5020738,"logger":"setup","msg":"start webhook server","listen_address":":9443"}
{"level":"info","ts":1661258810.5020816,"logger":"setup","msg":"starting anonymous reports"}
W0823 12:46:50.502112       1 client_config.go:617] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
{"level":"info","ts":1661258810.7947516,"logger":"setup","msg":"starting telemetry manager"}
{"level":"info","ts":1661258810.794883,"logger":"setup","msg":"starting manager"}
{"level":"info","ts":1661258810.7954898,"logger":"controller-runtime.webhook.webhooks","msg":"Starting webhook server"}
{"level":"info","ts":1661258810.7955565,"msg":"Starting server","kind":"health probe","addr":"[::]:8081"}
{"level":"info","ts":1661258810.7955763,"msg":"Starting server","path":"/metrics","kind":"metrics","addr":"127.0.0.1:8080"}
I0823 12:46:50.795735       1 leaderelection.go:248] attempting to acquire leader lease kong-system/a7feedc84.konghq.com...
[...]

PR #226, opened 2022-08-24
Improve status handling in reconciliation loops

mlavacca:
This PR aims at avoiding conflicts while updating resources. The following has been implemented:
* When the controlplane is deleted, the `clusterRole` and `ClusterRoleBinding` deletion leads to the end of the current loop. The object deletion will trigger the next loop.
* The `Controlplane` finalizers needed to remove cluster-wide resources are now removed through a patch operation instead of an update.
* In the Gateway controller, every `Gateway.Status` update is decided upon a comparison between the old gateway and the new one. Only if it is really needed, the update is performed.
* The gateway status updates have been replaced by patches.
* When setting the final `Ready` status of the Gateway, a unique Status update is now performed, instead of two subsequent updates.

**Which issue this PR fixes**
#203 

**PR Readiness Checklist**:

~- [ ] the `CHANGELOG.md` release notes have been updated to reflect significant changes~


controllers/controlplane_controller.go
```diff
@@ -110,24 +110,33 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		}
+		if deletions {
+			return ctrl.Result{}, nil // ClusterRoleBinding deletion will requeue
 		}
 
 		// now that ClusterRoleBindings are cleaned up, remove the relevant finalizer
-		if k8sutils.RemoveFinalizerInMetadata(&controlplane.ObjectMeta, string(ControlPlaneFinalizerCleanupClusterRoleBinding)) {
-			return ctrl.Result{}, r.Client.Update(ctx, controlplane) // Controlplane update will requeue
+		if k8sutils.RemoveFinalizerInMetadata(&newControlplane.ObjectMeta, string(ControlPlaneFinalizerCleanupClusterRoleBinding)) {
+			return ctrl.Result{}, r.Client.Patch(ctx, newControlplane, client.MergeFrom(controlplane)) // Controlplane update will requeue
```
randmonkey, line 125: Is there any benefit to use `Patch` instead of `Update`? To avoid conflicts?

mlavacca: Yes, conflicts shouldn't happen with Patch. Furthermore, it is idempotent (applying twice the same patch, the `resourceVersion` isn't touched).

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -71,58 +71,53 @@ func (r *GatewayReconciler) createControlPlane(
...
-		)
-		if err != nil {
-			return err
-		}
-
-		count := len(services)
-		if count > 1 {
-			return fmt.Errorf("found %d services for DataPlane currently unsupported: expected 1 or less", count)
-		}
+func (r *GatewayReconciler) ensureGatewayConnectivityStatus(ctx context.Context, gateway *gatewayDecorator, dataplane *operatorv1alpha1.DataPlane) (message string) {
```
randmonkey, line 74: Why do we want to change the function signature to return `string` here? I know the purpose here is to directly use the returned value as message in the condition, but IMO it is better to return an error here to annotate that something incorrect happened.

randmonkey (commented):
Generally good, but some minor comments remains.

controllers/gateway_controller_reconciler_utils.go
mlavacca: No specific reason. Since we'd use that error as a string, why not return a string? Just a matter of readability, but I can change the signature and return an error if you deem it :)

shaneutt (changes requested):
Looks great, a very thoughtful patch :+1: 

My comments are all minor, my only side notes are these:

- we should verify that we've performed manual testing before merging this PR
- if feasible we should extend our E2E tests: creating 5+ `Gateway` resources should result in no 409's in the pod logs
- we should look for any opportunity to extend our current tests, if I'm not mistaken we've added a condition where there's a connectivity error for the `Gateway`, correct? Can we have tests that verify this state?

mlavacca:
> we should verify that we've performed manual testing before merging this PR

I performed manual testing by deploying up to 10 Gateways at the same time and no error occurred. But I found out something curious: with that number of gateways deployed in parallel, the time needed by the Gateways to get ready rose above 3 minutes. I think this happens because there is a lack of resources in the cluster, but the deployments don't have requests set. The reconciliation is very fast, but the dataplane pods take a lot to get ready, and sometimes it can happen that they don't reply to the controlplane resulting in the controlplane crash. I think we can solve this issue by setting resource requests (and we should also set limits) in both controlplanes and dataplanes.

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -71,57 +71,52 @@ func (r *GatewayReconciler) createControlPlane(
...
-		for _, ip := range append(gatewayIPs, svc.Spec.ClusterIP) {
-			newAddresses = append(newAddresses, gatewayv1alpha2.GatewayAddress{
-				Type:  &ipaddrT,
-				Value: ip,
-			})
+	gatewayIPs := make([]string, 0)
+	if len(svc.Status.LoadBalancer.Ingress) > 0 {
+		ip := svc.Status.LoadBalancer.Ingress[0].IP
+		if ip == "" {
+			return errors.New("missing loadbalancer.ingress[0].ip")
```
pmalek, line 104: Small nit: maybe add service name to the error message?

controllers/gateway_controller_reconciler_utils.go
```diff
@@ -71,57 +71,52 @@ func (r *GatewayReconciler) createControlPlane(
...
+		gatewayIPs = append(gatewayIPs, ip) // TODO: handle hostnames https://github.com/Kong/gateway-operator/issues/24
+	}
 
-		gateway.Status.Addresses = newAddresses
-
-		k8sutils.SetReady(gateway)
-		return r.Client.Status().Update(ctx, gateway.Gateway)
+	newAddresses := make([]gatewayv1alpha2.GatewayAddress, 0, len(gatewayIPs))
+	ipaddrT := gatewayv1alpha2.IPAddressType
+	for _, ip := range append(gatewayIPs, svc.Spec.ClusterIP) {
```
pmalek, line 111: Hmm 🤔 this range param seems a little unconventional. Can we just `append` to `gatewayIPs` in a separate line? Wouldn't that bit a bit more readable? WDYT?

shaneutt:
> > we should verify that we've performed manual testing before merging this PR
> 
> I performed manual testing by deploying up to 10 Gateways at the same time and no error occurred. But I found out something curious: with that number of gateways deployed in parallel, the time needed by the Gateways to get ready rose above 3 minutes. I think this happens because there is a lack of resources in the cluster, but the deployments don't have requests set. The reconciliation is very fast, but the dataplane pods take a lot to get ready, and sometimes it can happen that they don't reply to the controlplane resulting in the controlplane crash. I think we can solve this issue by setting resource requests (and we should also set limits) in both controlplanes and dataplanes.
> 
> EDIT: I created an issue for that in #240

Sounds good, albeit there was an issue open which already was asking for memory and cpu limits, I commented on #240 about it.

> > if feasible we should extend our E2E tests: creating 5+ Gateway resources should result in no 409's in the pod logs
> 
> Considering that I replaced all the gateway status updates with patches, do you think this is something we should check in any case? I mean, theoretically, no conflict can happen with patches

We've changed the behavior of reconciliation in a relatively significant way. The main goal I have with this comment is that we should try and capture the intent of how things should be working in tests such that if someone made a PR that went against that prior intent (e.g. introduced a regression) we would get test feedback about this rather than silently allowing it through.

> > we should look for any opportunity to extend our current tests, if I'm not mistaken we've added a condition where there's a connectivity error for the Gateway, correct? Can we have tests that verify this state?
> 
> That status already existed, I changed only the way its update is performed. Since this is a status related to a transitory state, I see hard to test the existence of that status in the integration tests because it implies that either the cluster is not working properly (addresses are not assigned), or the operator is not working properly (the service hasn't been created). I think we should test this kind of stuff in the controller's unit tests.

Adding unit tests is always a good idea. As far as something more on the integration side of things I get that creating "broken" setups can be very challenging and time consuming, and you do have to balance that against the value you're adding. Use your best judgement here.

shaneutt (approved):

shaneutt (approved):
I think we should try and add some unit test coverage as previously discussed, then lgtm.

mlavacca:
> I think we should try and add some unit test coverage as previously discussed, then lgtm.

I'm working on adding a new e2e test that checks the operator logs. Do you prefer it to be added by a subsequent PR or do we wait for it before merging this one?

shaneutt:
> > I think we should try and add some unit test coverage as previously discussed, then lgtm.
> 
> I'm working on adding a new e2e test that checks the operator logs. Do you prefer it to be added by a subsequent PR or do we wait for it before merging this one?

If you're actively working on it, then ideally let's wait and get it in here.

mlavacca:
> > > I think we should try and add some unit test coverage as previously discussed, then lgtm.
> > 
> > 
> > I'm working on adding a new e2e test that checks the operator logs. Do you prefer it to be added by a subsequent PR or do we wait for it before merging this one?
> 
> If you're actively working on it, then ideally let's wait and get it in here.

I just realized it'd be way easier to merge this change before the e2e test because in the e2e tests we deploy the operator from main, and without having this change merged it's tricky to properly test it. I'll merge this one and open a new PR with the e2e test.

PR #227, opened 2022-08-24
fix(build): don't use temp dir for clien-gen

pmalek:

This PR aims to address using a temp dir when generating clientsets by using `client-gen`'s `--trim-path-prefix` flag.

This is basically a cherry-pick of https://github.com/Kong/kubernetes-ingress-controller/pull/2858


shaneutt (approved):

pmalek:
Tests failed. Race detected which will be fixed in https://github.com/Kong/gateway-operator/pull/218. The other failure might or might not be related. Rerunning.

PR #238, opened 2022-08-26
chore: objects in the gateway sample sorted

mlavacca:
Applying the file `config/samples/gateway.yaml` when the controller is running produces errors because of the objects sort: the `Gateway` should be created only after the `GatewayConfiguration`, otherwise the operator will reconcile the `Gateway` but the referenced `GatewayConfiguration` won't be found, and an error will be triggered.

**Which issue this PR fixes**

**PR Readiness Checklist**:


pmalek (approved):

shaneutt:
@mlavacca just so we're clear: this wasn't _stopping_ reconcilation from working properly correct? Even when the error was produced, it eventually became consistent? This is just a cosmetic change to make the examples run with less noise?

mlavacca:
> @mlavacca just so we're clear: this wasn't _stopping_ reconcilation from working properly correct? Even when the error was produced, it eventually became consistent? This is just a "mostly cosmetic" change to make the examples run with less noise?

Yep, the reconciliation of the gateway was performed in any case. This is only cosmetic to avoid that the first message a new user gets is an error.

PR #243, opened 2022-08-27
fix: remove networking.k8s.io DataPlane from PROJECT

pmalek:

It seems that the `DataPlane` in `PROJECT` file has 2 entries. 1 for a CRD from `domain:networking.k8s.io` and `group: gateway` and second one for a CRD from `domain:konghq.com` and `group:gateway-operator`. This PR removed the former since there is no `DataPlane` CRD in `networking.k8s.io`, `network` group.

This PR also changes the domains for `ControlPlane` and `DataPlane` from `konghq.com` to `gateway-opeartor.konghq.com` as that is the correct domain for those CRDs.


randmonkey:
I know the `PROJECT` file could be used for generating CRDs, but how is it used in our project? would you please add more detail in the comment of the PR, or create an issue for that.

shaneutt (approved):

shaneutt:
It's needed by `kubebuilder` / `operator-sdk` (built on `kubebuilder`) to generate files and build package bundles.

shaneutt (dismissed):

jrsmroz (approved):

PR #246, opened 2022-08-29
fix(dataplane) always set default envs

### LINKED ISSUE #245: env of deployment for dataplane could not be set when dataplane has env

### Current Behavior

Similar to #63, The default env could not be set if dataplane has set `spec.deploymentoptions.env`. Kong uses `postgres` as its default database, so this will make Kong fail to start if env is set in `spec.deploymentoptions.env` , but `KONG_DATABASE` not set.

### Expected Behavior

When I specify env in dataplane, the default envs are still set.

### Steps To Reproduce

Similar to #63, create a dataplane like this:
```
apiVersion: gateway-operator.konghq.com/v1alpha1
kind: DataPlane
metadata:
  creationTimestamp: "2022-08-29T09:47:14Z"
  generation: 1
  name: dp1
spec:
  env:
  - name: TEST_ENV
    value: before_update

```
The default envs  are not set, and the deployment could not be started, because `KONG_DATABASE` is not set,.
### Kong Gateway Operator version

_No response_

### Kubernetes version

_No response_

### Anything else?

_No response_

randmonkey:

Always set default envs of dataplane.

**Which issue this PR fixes**

fixes #245 

**PR Readiness Checklist**:


internal/utils/dataplane/config.go
```diff
@@ -73,12 +73,16 @@ var KongDefaults = map[string]string{
 // SetDataPlaneDefaults sets any unset default configuration options on the
 // DataPlane. No configuration is overridden. EnvVars are sorted
 // lexographically as a side effect.
-func SetDataPlaneDefaults(spec *operatorv1alpha1.DataPlaneDeploymentOptions) {
+// returns true if
```
pmalek, line 76: This comments looks unfinished 🤔

test/integration/dataplane_test.go
```diff
@@ -66,6 +77,22 @@ func TestDataplaneEssentials(t *testing.T) {
...
+	deployment := &deployments[0]
+
+	controllerContainer := k8sresources.GetPodContainerByName(
+		&deployment.Spec.Template.Spec, consts.DataPlaneProxyContainerName)
+	require.NotNil(t, controllerContainer)
+	envs := controllerContainer.Env
+	testEnvValue := getEnvValueByName(envs, "TEST_ENV")
+	require.Equal(t, "test", testEnvValue)
+	kongDatabaseEnvValue := getEnvValueByName(envs, consts.EnvVarKongDatabase)
+	require.Equal(t, "off", kongDatabaseEnvValue)
```
pmalek, line 94: This is coming from the default set of envs? Perhaps we comment that or add a message to the assertion?

internal/utils/dataplane/config.go
randmonkey: fixed.

test/integration/dataplane_test.go
randmonkey: added comments here.

pmalek (approved):

PR #247, opened 2022-08-29
fix: use gateway and controller related images from CSV

### LINKED ISSUE #102: dataplane and controlplane deployments should use Red Hat UBI images when operator is running UBI image

### Problem Statement

Currently, when user does not provide an image for a dataplane or a controlplane deployment, a default images is being used. The defaults are not based on Red Hat UBI image. 

One of the requirements for Red Hat certification is to base images on the Red Hat UBI base image. I believe it makes no sense to certify an operator image if it is going to pull non UBI images for deployments.

### Proposed Solution

- Operator uses the dedicated UBI images (if available) when the operator is running UBI itself.
- The decision can be made based on ENV variable.
- Because distroless and RedHat should be separate docker build targets, it's possible to set an ENV variable during build of a docker image.

### Additional information

_No response_

### Acceptance Criteria

- [ ] When operator is running Red Hat Universal Base Image, the default images used for controlplane and dataplane deployments are UBI too, if available.
- [ ] When operator is not running distroless image, the images used for controlplane and dataplane deployments are disrtroless or equivalent.
- [ ] User can still overwrite the images for dataplane and controlplane deployments.

jrsmroz:

Red Hat Certified operator should used KIC and Gateway images based on the Red Hat UBI. But the distrolless operator is not required to do so. This PR makes it possible to define a KIC and Gateway images in manager env vars. Then the env vars are defined in kustomization overlay specific for redhat bundle. This way anyone applying config manually will not drag in red hat UBI images, but the bundle will correctly include UBI images.

**Which issue this PR fixes**

Fixes #102 

**PR Readiness Checklist**:


jrsmroz:
OK, maybe make  I did something wrong here? `make bundle` removes the `RELATED_IMAGE_*` envs. I need to double check how those should be added.

jrsmroz:
So right now the images are global. I would like them to be only for Red Hat bundle. Any way to achieve this @rainest ?

jrsmroz:
So, I've used the kustomize overlay for a bundle to specify UBI imaged.

Makefile
```diff
@@ -261,6 +261,13 @@ bundle: manifests kustomize
 	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle $(BUNDLE_GEN_FLAGS)
 	$(OPERATOR_SDK) bundle validate ./bundle
 
+.PHONY: bundle-redhat-certified
+bundle-redhat-certified: manifests kustomize
+	$(OPERATOR_SDK) generate kustomize manifests -q
+	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+	$(KUSTOMIZE) build config/redhat-certified | $(OPERATOR_SDK) generate bundle $(BUNDLE_GEN_FLAGS)
+	$(OPERATOR_SDK) bundle validate ./bundle
```
pmalek, line 269: This looks pretty much identical to `bundle` target with the difference of kustomize dir being used. How about we create a "generic" `_bundle` (or using a different name) generic target and use a param for that directory name (similarly as it's being done with `docker.build` and `_docker.build`?

jrsmroz: Yes, good that you noticed. my main concern is whether the operator hub bundle should use UBI images? If so, maybe we need multiple bundles directories.

jrsmroz: Regarding the generic target, if we decide we need multiple targets it's good idea. Otherwise I will remove the `bundle-redhat-certified`. However I think it may be a way to keep all those redhat specific annotations and so on. Hence, it can even be done in bridge repo.

pmalek: > maybe we need multiple bundle directories.

Then we could easily add a `BUNDLE_DIR` param. My point here is that we're sort of juggling those manifests here and there (https://github.com/Kong/redhat-certified-operators) but for the time being this is the "source of truth" for now.

Whatever we decide later on (moving operator-sdk related stuff to a separate repo) can be improved by keeping this as intended in the long run. Since for now we don't have the machinery (nor the concrete plan) for moving the stuff to a separate repo I suggest we add that parametrized target here which could then be reused in other repos if need be.

jrsmroz: Ok, changed as suggested.

config/redhat-certified/manager_config_patch.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  template:
+    spec:
+      containers:
+      - name: manager
+        env:
+        - name: RELATED_IMAGE_KONG
+          value: registry.connect.redhat.com/kong/kong:2.8.1
+        - name: RELATED_IMAGE_KONG_CONTROLLER
+          value: registry.connect.redhat.com/kong/kong-ingress-controller:2.5.0-redhat
```
pmalek, line 15: I wonder what would be the best approach to manage this, I mean GW and KIC versions that we support as of `main`.

I believe we should tackle that problem "globally" for the whole repo, e.g. store those in a config file and use it during every code generation step that we run.

I'd agree on pushing this to a separate PR but we'd need an issue for that. Do you mind creating or do you want me to do it?

controllers/controlplane_controller_utils.go
```diff
@@ -200,6 +201,9 @@ func generateNewDeploymentForControlPlane(controlplane *operatorv1alpha1.Control
 		if controlplane.Spec.Version != nil {
 			controlplaneImage = fmt.Sprintf("%s:%s", controlplaneImage, *controlplane.Spec.Version)
 		}
+	} else if relatedKongControllerImage := os.Getenv("RELATED_IMAGE_KONG_CONTROLLER"); relatedKongControllerImage != "" {
+		// RELATED_IMAGE_KONG_CONTROLLER is set by the operator-sdk when building the operator bundle.
+		controlplaneImage = relatedKongControllerImage
```
pmalek, line 206: So this seems a little bit "magical" behavior. Basically what this will do is it will be used as fallback when no container image is specified but when that particular env is present 🤔 

I'd prefer to not have to put this type of "special case" code in here. OTOH I don't have a ready solution for this at this moment.

jrsmroz: If the image us specified in user provided configuration (1st if), we're not in the position to use something different.

config/redhat-certified/manager_config_patch.yaml
jrsmroz: The way I thought about this, was treating `redhat-certified/config/manager_config_patch` s source of truth for the supported images.

pmalek: But I mean: the image versions ( e.g. 2.8 for GW and 2.5 for KIC ) will be the same for redhat flavor of the operator and for the other flavors right? I'd suggest to keep those particular values somewhere for **all** flavors of the operator that we might end up releasing the same. Otherwise it might end up as chaos of supported versions, and a mess to maintain.

controllers/controlplane_controller_utils.go
jrsmroz: Ref : https://docs.openshift.com/container-platform/4.11/operators/admin/olm-restricted-networks.html
> While OLM can manage Operators from local sources, the ability for a given Operator to run successfully in a restricted network still depends on the Operator itself. The Operator must:List any related images, or other container images that the Operator might require to perform their functions, in the relatedImages parameter of its ClusterServiceVersion (CSV) object.Reference all specified images by a digest (SHA) and not by a tag.

jrsmroz: But we may move that logic to place where the `controlplane.Spec` is being created.

config/redhat-certified/manager_config_patch.yaml
jrsmroz: Oh, I get it. Well, this is ideal situation and it should be like that. Let's create an issue to track it.

jrsmroz: https://github.com/Kong/gateway-operator/issues/258

controllers/controlplane_controller_utils.go
shaneutt: I think @pmalek is right to take pause: with this change we're effectively adding to our public interface "redhat-isms" and in general conforming to some standard like that is something we should have consensus on and have a good reason beyond simple convenience.

That said, this definitely seems to be the shortest path to ground: we can get `-redhat` builds for bundles OOTB with minimal effort, and in general unless someone decides to use this exact environment variable, it wont change the global defaults.

I see two alternatives in the immediate:

One alternative would seem to be to make our own standard environment variables / flags for controlling this default that doesn't conform to the `RELATED_IMAGE_*` style provided by `operator-sdk`, and then adding some hook somewhere to convert what the `make bundle` provides into our own flag. This will be somewhat time consuming and with gains that I don't see as being super strong.

Another alternative I think might have some promise would be to use a Go build tag to provide alternative builds. There are multiple potential downsides here however: using build tags in this way can come with some significant maintenance costs, and can make testing trickier. I don't think we should take this path just for this one var, we should wait until we have _super significant_ redhat-specific logic that's needed for compliance there, and even then we should still take pause and consider further the amount of extra maintenance that is potential.

Overall, I see both sides but given that we have a lot of `alpha` maturity runway, and plenty of time to change this if we decide we don't want it: I recommend we create a follow-up issue to revisit this and any similar "redhat-isms" that we may take on between now and a time before beta as a spike to re-evaluate these things and weigh and balance the alternatives. We document that here as a TODO but then we continue forward with what we have so we can get the first alpha release out with bundles that work right on other platforms. LMKWYT?

jrsmroz: The third alternative is to embedded that into the container via env var, but it's no different than using env var from the bundle.

Forth alternative is using `//go:embed` from within docker builds (red hat has it's own docker build).

Overall I believe it does not matte what we choose. It's undocumented interface, a subject to change at anytime. Users are provided with proper interface do accomplish what they want (changing the image used) anyway.

shaneutt: So is that a "yes" that you think having a check-in later (but continuing with this as-is for the purposes of this PR) is fine? Or did you want to do something different?

jrsmroz: For me - check in now. I want to finish the bundle certification related stuff asap.

shaneutt: Yes, that's what I was saying in my first comment here: I think we should follow up (with an issue) but that we should merge this as is for now and do that follow up at a later time (before beta).

jrsmroz: ok

pmalek (dismissed):
👍 

Ideally, if possible, I'd suggest to leave the link to the created issue (about `RELATED_*` env vars) next to where they are use in the code.

jrsmroz:
> 👍
> 
> Ideally, if possible, I'd suggest to leave the link to the created issue (about `RELATED_*` env vars) next to where they are use in the code.

Added, but it discarded the approvals.

pmalek (approved):

PR #252, opened 2022-08-29
feat(kic-role-generator): use KIC's submodule for generation

pmalek:

This PR adds KIC as submodule of this repo which is then used e.g. in `make generate.rbacs` target.

The aim of this is to prevent `git clone` on each invocation of `make generate.rbacs`


hack/generators/kic-role-generator/main.go
```diff
@@ -41,21 +42,41 @@ func init() {
 }
 
 func main() {
-	defer cleanup()
+	// c := git.Submodule{}
```
shaneutt, line 45: we should delete this leftover comment, or do something if there's something missing here.

pmalek: Addressed. Thanks!

shaneutt (approved):

pmalek:
I need to add https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/workflows/check_pr_labels.yaml to this repo here 😅

pmalek:
After having a discussion with @mlavacca on slack we believe that this might be a relatively high cost for such a little use case hence there's a counter proposal to make this a "permanent" checked out repo in rbac generator dir, added to `.gitignore`.

WDYT @Kong/team-k8s ?

shaneutt:
> After having a discussion with @mlavacca on slack we believe that this might be a relatively high cost for such a little use case hence there's a counter proposal to make this a "permanent" checked out repo in rbac generator dir, added to `.gitignore`.
> 
> WDYT @Kong/team-k8s ?

How would this be instrumented?

pmalek:
> > After having a discussion with @mlavacca on slack we believe that this might be a relatively high cost for such a little use case hence there's a counter proposal to make this a "permanent" checked out repo in rbac generator dir, added to `.gitignore`.
> > WDYT @Kong/team-k8s ?
> 
> How would this be instrumented?

Basically this would work the same way as it works now on `main` but with the twist that the check out dir is in `hack/generators/kic-role-generator/kubernetes-ingress-controller/` and it doesn't get removed after this script finishes. (unless instructed otherwise via flags)

shaneutt:
> > > After having a discussion with @mlavacca on slack we believe that this might be a relatively high cost for such a little use case hence there's a counter proposal to make this a "permanent" checked out repo in rbac generator dir, added to `.gitignore`.
> > > WDYT @Kong/team-k8s ?
> > 
> > 
> > How would this be instrumented?
> 
> Basically this would work the same way as it works now on `main` but with the twist that the check out dir is in `hack/generators/kic-role-generator/kubernetes-ingress-controller/` and it doesn't get removed after this script finishes. (unless instructed otherwise via flags)

Have you considered sub-moduling the KIC? What do you think are the pros/cons?

pmalek:
> > > > After having a discussion with @mlavacca on slack we believe that this might be a relatively high cost for such a little use case hence there's a counter proposal to make this a "permanent" checked out repo in rbac generator dir, added to `.gitignore`.
> > > > WDYT @Kong/team-k8s ?
> > > 
> > > 
> > > How would this be instrumented?
> > 
> > 
> > Basically this would work the same way as it works now on `main` but with the twist that the check out dir is in `hack/generators/kic-role-generator/kubernetes-ingress-controller/` and it doesn't get removed after this script finishes. (unless instructed otherwise via flags)
> 
> Have you considered sub-moduling the KIC? What do you think are the pros/cons?

Well this PR adds KIC as submodule 😅 

Cons 🤔 submodules? 😅 

But in all seriousness, the idea was to drop adding the submodule and use just a plain checked out repo due to possible headaches in the future related to use of submodules. For that simple and small use case as rbac manifests generation is I believe a regular cloned repo "managed" by this script might do the trick and not cause problems.

What's your take on this @shaneutt ?

shaneutt:
> > > > > After having a discussion with @mlavacca on slack we believe that this might be a relatively high cost for such a little use case hence there's a counter proposal to make this a "permanent" checked out repo in rbac generator dir, added to `.gitignore`.
> > > > > WDYT @Kong/team-k8s ?
> > > > 
> > > > 
> > > > How would this be instrumented?
> > > 
> > > 
> > > Basically this would work the same way as it works now on `main` but with the twist that the check out dir is in `hack/generators/kic-role-generator/kubernetes-ingress-controller/` and it doesn't get removed after this script finishes. (unless instructed otherwise via flags)
> > 
> > 
> > Have you considered sub-moduling the KIC? What do you think are the pros/cons?
> 
> Well this PR adds KIC as submodule sweat_smile
> 
> Cons thinking submodules? sweat_smile
> 
> But in all seriousness, the idea was to drop adding the submodule and use just a plain checked out repo due to possible headaches in the future related to use of submodules. For that simple and small use case as rbac manifests generation is I believe a regular cloned repo "managed" by this script might do the trick and not cause problems.
> 
> What's your take on this @shaneutt ?



> > > > > After having a discussion with @mlavacca on slack we believe that this might be a relatively high cost for such a little use case hence there's a counter proposal to make this a "permanent" checked out repo in rbac generator dir, added to `.gitignore`.
> > > > > WDYT @Kong/team-k8s ?
> > > > 
> > > > 
> > > > How would this be instrumented?
> > > 
> > > 
> > > Basically this would work the same way as it works now on `main` but with the twist that the check out dir is in `hack/generators/kic-role-generator/kubernetes-ingress-controller/` and it doesn't get removed after this script finishes. (unless instructed otherwise via flags)
> > 
> > 
> > Have you considered sub-moduling the KIC? What do you think are the pros/cons?
> 
> Well this PR adds KIC as submodule sweat_smile
> 
> Cons thinking submodules? sweat_smile
> 
> But in all seriousness, the idea was to drop adding the submodule and use just a plain checked out repo due to possible headaches in the future related to use of submodules. For that simple and small use case as rbac manifests generation is I believe a regular cloned repo "managed" by this script might do the trick and not cause problems.
> 
> What's your take on this @shaneutt ?

Sorry, I miss-worded my previous comment (updated above). What I was looking for was for us to reconsider submoduling: can we actually enumerate the problems we think will come from using a submoduling approach?

shaneutt:
We have plenty of time before we'll be in beta, and therefore plenty of time to rework this if we run into pain points. I think we should continue as-is and perhaps create an issue to remind ourselves to check in on it in the coming months if you're still feeling tentative.

shaneutt (approved):

PR #255, opened 2022-08-30
fix: add kustomize samples for bundle generation

pmalek:

This PR adds samples `ControlPlane` and `DataPlane` samples as well as more real realistic sample for `GatewayConfiguration`.

It also regenerates the bundle manifests via `make bundle`.

Thanks to this patch the following errors during bundle generation disappear:

```
WARN[0000] ClusterServiceVersion validation: [OperationFailed] provided API should have an example annotation
WARN[0000] ClusterServiceVersion validation: [OperationFailed] provided API should have an example annotation
INFO[0000] Creating bundle.Dockerfile
INFO[0000] Creating bundle/metadata/annotations.yaml
INFO[0000] Bundle metadata generated suceessfully
<DIR>/bin/operator-sdk bundle validate ./bundle
WARN[0000] Warning: Value gateway-operator.konghq.com/v1alpha1, Kind=ControlPlane: provided API should have an example annotation
WARN[0000] Warning: Value gateway-operator.konghq.com/v1alpha1, Kind=DataPlane: provided API should have an example annotation
```

### Open questions

Are those added samples representative enough? Or do we want to provide different ones instead?


config/samples/gateway.yaml
```diff
@@ -5,13 +5,12 @@ metadata:
...
-    - name: TEST_VAR
-      value: TEST_VAL
+    containerImage: kong
+    version: "2.8"
   controlPlaneDeploymentOptions:
     env:
-    - name: TEST_VAR
-      value: TEST_VAL
+    - name: CONTROLLER_PUBLISH_SERVICE
+      value: kong-system/kong-proxy
```
shaneutt, line 13: I wouldn't use this field, specifically, as an example. This is a field that the KGO should be managing in 99.99% of cases.

shaneutt (changes requested):

jrsmroz:
`bundle/manifests/operator.clusterserviceversion.yaml` is generated. Will that change be overwritten after running `make bundle`?

pmalek:
> `bundle/manifests/operator.clusterserviceversion.yaml` is generated. Will that change be overwritten after running `make bundle`?

`make bundle` will take kustomize files from `config/manifests` which includes `../samples` which in turn contains files that I changes in this PR. This in turn is taken into account when generating `clusterserviceversion.yaml`

So to answer your question: I've regenerated the files so that now `make bundle` doesn't produce a diff as of this PR.

config/samples/gateway.yaml
pmalek: Sure, anything else that I could put there instead?

shaneutt: Perhaps debug logging, as that's something a user would actually do if trying to diagnose a potential KIC issue. Similarly, enabling profiling and configdumps might be good ones in the same motif.

config/samples/dataplane.yaml
```diff
@@ -2,3 +2,7 @@ apiVersion: gateway-operator.konghq.com/v1alpha1
 kind: DataPlane
 metadata:
   name: dataplane-example
+spec:
+  env:
+  - name: KONG_PROXY_LISTEN
+    value: "0.0.0.0:8000, 0.0.0.0:8443 ssl http2"
```
shaneutt, line 8: Sorry I didn't catch this the first time. This one is iffy too, it's something the KGO should be managing. Perhaps try log level for this one as well, as that's something more suited to the user managing. It's been an ongoing conversation how much "raw access" we want to provide, today we simply provide full access, but we'll need to keep reconsidering how users should be accessing these fields as we move further into alpha releases. My inclination is that at some point we should identify and wrap _every_ field we think a user should use as a clean part of the spec, and still make the raw access available but simply make it less prominent and document it as here be dragons.

pmalek: Got it, changed to `KONG_LOG_LEVEL`.

shaneutt (approved):

PR #257, opened 2022-08-30
chore: don't delete cluster when provided

mlavacca:
When the cluster is provided through the env variable, do not delete it when an error occurs.

**Which issue this PR fixes**

**PR Readiness Checklist**:


pmalek:
🤔 While useful for debugging, this can end up being expensive if e2e tests are run e.g. on GKE. WDYT?

mlavacca:
> 🤔 While useful for debugging, this can end up being expensive if e2e tests are run e.g. on GKE. WDYT?

When tests are run in the CI that env var should be never set. Furthermore, for now, if the test is successful the cluster is not deleted, it happens only when an error occurs.

pmalek:
> > 🤔 While useful for debugging, this can end up being expensive if e2e tests are run e.g. on GKE. WDYT?
> 
> When tests are run in the CI that env var should be never set. Furthermore, we are inconsistent in our behavior: if the test succeeds, the cluster is left alive, otherwise, is deleted.

Yes, but I was trying to say is that when test will run on GKE or any other cloud provider (which I guess is the plan? 🤔 ) then a failed test would result in dangling resources (cluster running in the cloud). What will we do in that case? Go and manually delete it?

mlavacca:
> > > 🤔 While useful for debugging, this can end up being expensive if e2e tests are run e.g. on GKE. WDYT?
> > 
> > 
> > When tests are run in the CI that env var should be never set. Furthermore, we are inconsistent in our behavior: if the test succeeds, the cluster is left alive, otherwise, is deleted.
> 
> Yes, but I was trying to say is that when test will run on GKE or any other cloud provider (which I guess is the plan? 🤔 ) then a failed test would result in dangling resources (cluster running in the cloud). What will we do in that case? Go and manually delete it?

Yes, that is what we do when the test succeeds. The env var `KONG_TEST_CLUSTER` exists for providing an already existing cluster that we don't want to be deleted when the test ends, no matter the outcome.

shaneutt (approved):
This makes sense to me, this would be more consistent with how KIC does this. As for the GKE argument: we have a cron cleanup job to do this: https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/workflows/cleanup.yaml because we can't _fully_ rely on the code to clean itself up.

pmalek:
> This makes sense to me, this would be more consistent with how KIC does this. As for the GKE argument: we have a cron cleanup job to do this: https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/workflows/cleanup.yaml because we can't _fully_ rely on the code to clean itself up.

We'd need to retrofit https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/workflows/cleanup.yaml into this repo, right? I suggest we create an issue for this so it's not lost.

shaneutt:
> > This makes sense to me, this would be more consistent with how KIC does this. As for the GKE argument: we have a cron cleanup job to do this: https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/workflows/cleanup.yaml because we can't _fully_ rely on the code to clean itself up.
> 
> We'd need to retrofit https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/workflows/cleanup.yaml into this repo, right? I suggest we create an issue for this so it's not lost.

If we use KTF and don't change anything about how it creates a GKE cluster, no: it's label based and the default labels handle it so the KIC cron would clear it (even if that is a bit weird to have that repository be the one responsible, particularly). I suppose what would probably be better is once we add real GKE support here we move that cron to the KTF repo since KIC won't be the only user anymore? :thinking:

jrsmroz (approved):

PR #264, opened 2022-08-31
feat: print version at startup

### LINKED ISSUE #201: Log the version and commit hash on startup

### Problem Statement

I was writing up a bug report for https://github.com/Kong/gateway-operator/issues/200 (that runs `ghcr.io/kong/gateway-operator:main`) and realized that there's no obvious way (known to me at least) to identify the commit hash of the build I was using.

### Proposed Solution

On KGO startup, log the `internal/metadata` vars (`Release`, `Repo`, `Commit`).

### Additional information

KIC does this:

https://github.com/Kong/kubernetes-ingress-controller/blob/04fcf2a5ec00df5c9418c2bf6e612ac893db47c8/internal/manager/run.go#L41

### Acceptance Criteria

- [ ] As a user of KGO, I can read from logs the commit hash and the version number of KGO.

pmalek:

This PR adds version printing on startup.

Additionally it fixes the logger setup at the beginning of `manager.Run()`. Before this patch logger wasn't configured at the very beginning of the function and any log entries that were supposed to be printed were discarded because at that point the logger wasn't configured.

After [the recommendation](https://github.com/Kong/gateway-operator/pull/264#pullrequestreview-1091728983) from @shaneutt I ended up adding a reusable build step which was used then subsequently to test the image.

This also adds `scripts/verify-version.sh` that verifies the version (that is piped in to script's stdin) with a one that it generates ( based on the logic from the operator ).

Exemplar run of CI testing the image in my fork: https://github.com/pmalek/gateway-operator/runs/8138892668?check_suite_focus=true#step:6:1

**Which issue this PR fixes**

Fixes #201

### Open questions

Do we want to run the image version check for PRs as well? As of now (this PR) this only runs on builds for pushes against `main`.


shaneutt (dismissed):
LGTM, but while we're in here I think we should add a test that verifies the metadata is getting populated in the container build, e.g. the test does:

1. builds the image in the same way the [build workflow](https://github.com/Kong/gateway-operator/blob/main/.github/workflows/build.yaml) does
2. verifies that the version is being logged correctly

The reason I point this out is that we actually had multiple regressions of this with the [KIC](https://github.com/kong/kubernetes-ingress-controller) due to refactoring and moving code around and I would like to avoid regressions here as we'll almost definitely be doing major cleanups here too before beta.

pmalek:
> 

I'm all in for for that but in order for this test to make sense I suggest to make a custom workflow out of https://github.com/Kong/gateway-operator/blob/33dfe6af016402fc8e77696d16cd22dc7addbd4e/.github/workflows/build.yaml#L29-L45 and call it both during release build and during the test build so that we use the same mechanism with the same parameter set.

Otherwise we might end up in a situation where we're testing something else than what we're releasing.

Unless someone's got a better idea for this.

shaneutt:
> > 
> 
> I'm all in for for that but in order for this test to make sense I suggest to make a custom workflow out of
> 
> https://github.com/Kong/gateway-operator/blob/33dfe6af016402fc8e77696d16cd22dc7addbd4e/.github/workflows/build.yaml#L29-L45
> and call it both during release build and during the test build so that we use the same mechanism with the same parameter set.
> 
> Otherwise we might end up in a situation where we're testing something else than what we're releasing.
> 
> Unless someone's got a better idea for this.

As long as it `does-the-thing.py` the actions job approach seems fine to me :+1:

pmalek:
> As long as it `does-the-thing.py` the actions job approach seems fine to me 👍

Of course. My main concern was that whatever we'll use in release will diverge (at some point, without anyone noticing) and we won't test what needs testing.

I'll try to cook something up.

pmalek:
@shaneutt PTAL. I've added some magic to test what was added in this PR.

shaneutt (approved):
So this setup feels like it's gonna end up becoming perhaps slightly esoteric, but it's an effective means to reduce regressions for now. I say we roll with this for the moment, patch as needed, or replace as needed after we've spent some time with it.

PR #269, opened 2022-08-31
logging e2e test

mlavacca:
Addition of a e2e test that:
1. creates 5 gateways sequentially
2. wait for all the gateways to be ready
3. deletes all the 5 gateways sequentially
4. wait for all the sub-resources are correctly deleted
5. check the gateway operator error message: no error must occur (apart those set in the map of the allowed error messages)

**Which issue this PR fixes**

This PR is a follow-up of #226, where the resources status handling has been improved so that no conflict should happen now.

**PR Readiness Checklist**:


Makefile
```diff
@@ -324,6 +324,10 @@ test.integration:
 test.e2e:
 	GOFLAGS="-tags=e2e_tests" go test -race -v ./test/e2e/...
 
+.PHONY: test.logs-e2e
+test.logs-e2e:
+	GOFLAGS="-tags=logs_e2e_tests" go test -race -v ./test/e2e/...
```
shaneutt, line 329: It's not clear to me why this E2E test has become an entirely separate thing, is this because our E2E testing suite currently provides a "default" cluster where in this case we want our own? Because I think that's a problem we've been needing to solve with the E2E test suite in general, and I would hate to start down this path of adding more build tags and more separation when what we really want is for each E2E test to just build and manage its own cluster :thinking:

shaneutt (changes requested):

Makefile
mlavacca: yes, the issue here is that we need a fresh cluster with a fresh operator to run this test. We need this because we read all the operator logs and check that no error occurs. In a dirty env in which other tests have already run (such as the webhook test), we end up checking also the log lines produced by those tests and eventually make the test fail because of errors triggered by other tests (maybe on purpose). I don't like this approach either, but it's the only one I found that fits our e2e env.

shaneutt: How do you feel about taking this time to rework the E2E environment so that it matches a bit closer to KIC? e.g. drop the current cluster setup and instead make each test responsible for it's own cluster and/OR combining tests which are compatible?

mlavacca: Sure, I created an issue for this: https://github.com/Kong/gateway-operator/issues/279

test/e2e/operatorLogs_test.go
```diff
@@ -0,0 +1,273 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gatewayclient "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned"
+
+	gatewayutils "github.com/kong/gateway-operator/internal/utils/gateway"
+)
+
+const (
+	// the total number of gateways that are created and deleted one after the other
```
shaneutt, line 33:
```suggestion
	// parallelGateways is the total number of gateways that are created and deleted one after the other
```

test/e2e/operatorLogs_test.go
```diff
@@ -0,0 +1,273 @@
...
+	parallelGateways = 5
+	// gatewayReadyTimeLimit is the maximum amount of time to wait for a
+	// supported Gateway to be fully provisioned and marked as Ready by the
+	// gateway controller.
+	gatewayReadyTimeLimit = time.Minute * 3
```
shaneutt, line 38: We have this concept in the integration test suite as well, don't we? I wouldn't be opposed to starting to collect these things (especially the ones that can be shared between the integration and e2e test suites, because they should be the same) somewhere like `internal/utils/test/` or something like that :thinking:

test/e2e/operatorLogs_test.go
```diff
@@ -0,0 +1,273 @@
...
+	defer func() {
+		if t.Failed() {
+			output, err := cleaner.DumpDiagnostics(ctx, t.Name())
+			t.Logf("%s failed, dumped diagnostics to %s", t.Name(), output)
+			assert.NoError(t, err)
+		}
+		assert.NoError(t, cleaner.Cleanup(ctx))
+	}()
+
+	// get the gateway operator pod
```
shaneutt, line 76:
```suggestion
	t.Log("finding the Pod for the Gateway Operator")
```

test/e2e/operatorLogs_test.go
```diff
@@ -0,0 +1,273 @@
...
+		}
+
+		t.Log("verifying the networkpolicy is deleted")
+		require.Eventually(t, Not(gatewayNetworkPoliciesExist(t, ctx, &gateway, clients.mgrClient)), time.Minute, time.Second)
+	}
+}
+
+// TODO: the helpers below are copy-paste from the integration test folder.
+// We should create an util package under tests that can be used by both
+// integration and e2e tests
```
shaneutt, line 189: This fits in with a comment I made above, let's just put this stuff in `internal/utils/test/` wdyt?

shaneutt (changes requested):
Mostly looks good, just a couple minor comments.

test/e2e/operatorLogs_test.go
mlavacca: Sure, I'll do this refactoring

test/e2e/operatorLogs_test.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+
+		t.Log("verifying the networkpolicy is deleted")
+		require.Eventually(t, Not(gatewayNetworkPoliciesExist(t, ctx, &gateway, clients.MgrClient)), time.Minute, time.Second)
+	}
+}
+
+// TODO: the helpers below are copy-paste from the integration test folder.
+// We should create an util package under tests that can be used by both
+// integration and e2e tests
```
shaneutt, line 190: Have we completed consolidating the helpers? Would seem we should remove this comment if so?

mlavacca: Yep, I rebased the PR quickly and overlooked removing this comment and the comment below. It should be fine now.

shaneutt (approved):

PR #276, opened 2022-09-01
chore: put regular and redhat-certified bundles into separate dirs

pmalek:

This PR tries to tackle the problem of differences between generated bundles due to different configs (e.g. redhat certified bundle needing related images patch).

This also:

- moves the `bundle.Dockerfile` that was present at the root of the repository into `bundle/regular/` dir
- adds 1 more Dockerfile, so that now we have 2:
  - `bundle/regular/bundle.Dockerfile` for "regular" builds
  - `bundle/redhat-certified/bundle.Dockerfile` for redhat certified builds, that have the `RELATED_IMAGE_*` env variables injected

~~This is sort of incomplete because `bundle.Dockerfile` - seems like not configurable in `operator-sdk generate bundle ...` - will get regenerated and changed depending on which bundle is being generated (either `bundle` or `bundle-redhat-certified`)~~

~~This needs to be addressed somehow (e.g. by adding manual renaming commands in `Makefile`?)~~


config/redhat-certified/related_images.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: kong-gateway-operator.v0.0.0
+  namespace: placeholder
+spec:
+  relatedImages:
+  - image: registry.connect.redhat.com/kong/kong:2.8.1
+    name: kong
+  - image: registry.connect.redhat.com/kong/kong-ingress-controller:2.5.0-redhat
```
jrsmroz, line 10: If the `RELATED_IMAGE_` env vars are used, the `spec.relatedImages` are updated automatically.
However, it seems the certification pipeline want also the rbac proxy and operator images to be included in `spec.relatedImages`

config/redhat-certified/related_images.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: kong-gateway-operator.v0.0.0
```
jrsmroz, line 4: I propose we start using v0.0.1. I actually made other PR to bump that.

config/redhat-certified/related_images.yaml
pmalek: You're right. I've dropped the `related_images.yaml` patch. This also nullifies the usage of specific version in said patch 👍

config/redhat-certified/related_images.yaml
pmalek: Explained in the other comment. This is not needed anymore.

Makefile
```diff
@@ -34,14 +34,19 @@ endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 IMAGE_TAG_BASE ?= ghcr.io/kong/gateway-operator
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
-BUNDLE_GEN_FLAGS ?= -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
+BUNDLE_GEN_FLAGS ?= --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
```
jrsmroz, line 37: A question: does that define a bundle version or it's used for defining the operator version.

jrsmroz (dismissed):

Makefile
pmalek: This will set `metadata.name`'s version and `spec.version` in CSV.

pmalek: Does that answer your question or do you have any other concerns about it?

jrsmroz (approved):

PR #277, opened 2022-09-01
fix: add more metadata used in redhat certified operator

jrsmroz:

- Adds Kong icon
- Adds links to github repository
- Adds maintainers contact info

**Which issue this PR fixes**

**PR Readiness Checklist**:


config/manifests/bases/kong-gateway-operator.clusterserviceversion.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   annotations:
     alm-examples: '[]'
     capabilities: Basic Install
-  name: kong-gateway-operator.v0.0.0
+  name: kong-gateway-operator.v0.0.1
```
pmalek, line 7: I'm wondering if we wouldn't prefer to leave this as `0.0.0` because it's a base template. We'll be overriding that upon release anyway. WDYT?

config/manifests/bases/kong-gateway-operator.clusterserviceversion.yaml
```diff
@@ -52,10 +52,13 @@ spec:
...
+    url: https://github.com/kong/gateway-operator
+  maintainers:
+  - email: support@konghq.com
+    name: Kong Inc.
   maturity: alpha
   provider:
     name: Kong Inc.
     url: https://konghq.com
-  version: 0.0.0
+  version: 0.0.1
```
pmalek, line 64: Ditto here

config/manifests/bases/kong-gateway-operator.clusterserviceversion.yaml
jrsmroz: You mean to always left it at v0.0.0 or till first release? 

I'm not strongly opiated here. That depend if we're going to publish separate version on operator-hub, certified operators and in future marketplane. The marketplace may need special treatment.

pmalek: Not exactly. What I'm saying is that we could put `v0.0.0` here, as it's not used anywhere. It's just a template. We're overriding this in the generation step anyway: https://github.com/Kong/gateway-operator/blob/33dfe6af016402fc8e77696d16cd22dc7addbd4e/Makefile#L37.

And the reason I'm proposing this is that this will be one less thing to maintain and it will be obvious as new versions come along that it's not something that should be updated (given that we'll have e.g. version `v0.2.15` in other manifests and `v0.0.0` here).

jrsmroz: Ahh, ok.

pmalek (approved):

PR #278, opened 2022-09-01
Logging improvement

### LINKED ISSUE #189: Restructure logging

### Problem Statement

As of now, our logging doesn't help in providing context about what's happening in the controllers but is limited to a bullet list of actions that are executed, without focusing on the important aspect of those actions: the outcome. We should improve our debug logging in a way that focus is given on what happened, instead of what's going to happen. 
An example with more details follows:

```go
	debug(log, "exposing DataPlane deployment via service", dataplane)
	created, dataplaneService, err := r.ensureServiceForDataPlane(ctx, dataplane)
	if err != nil {
		return ctrl.Result{}, err
	}
	if created {
		return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil
	}
```

would be more meaningful and helpful if replaced with something similar:

```go
	created, dataplaneService, err := r.ensureServiceForDataPlane(ctx, dataplane)
	if err != nil {
		return ctrl.Result{}, err
	}
	if created {
	        debug(log, "DataPlane deployment exposed via service", dataplane)
		return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil
	}
```

A lot of further helpful information could also be logged from within the reconciler utils functions, to provide more context about what's happening in the controller. By using this approach, we'd replace the massive amount of redundant log lines with a little and more meaningful amount.

### Proposed Solution

To replace the "bullet-list" logging style with a more descriptive logging style that focuses on what's happened in the previous step instead of what is going to happen in the next reconciliation step.

### Acceptance Criteria

- [x] Each log line with debug verbosity in KGO that logs about an action that is going to happen should be either:
  * replaced by the same message but with a lower verbosity
  * removed
- [x] Every time an important successful event (resources are created, status is updated, etc.) for the reconciliation loop happens, we should either:
  * log it with debug verbosity, or
  * log it with info verbosity
- [x] every time an error occurs and that error is not used to trigger a requeue through `return crtrl.Result{}, err`, we should log that error with debug verbosity.

mlavacca:
Signed-off-by: Mattia Lavacca <lavacca.mattia@gmail.com>

log improvement to filter what events are happening in the cluster.

**Which issue this PR fixes**
Fixes #189 

**PR Readiness Checklist**:


controllers/gateway_controller.go
```diff
@@ -188,59 +197,68 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-	debug(log, "gateway reconciled", gateway)
+	debug(log, "reconciliation complete for Gateway resource", gateway)
 	return ctrl.Result{}, nil
 }
 
-func (r *GatewayReconciler) provisionDataPlane(ctx context.Context, log logr.Logger, gateway *gatewayDecorator, gatewayConfig *operatorv1alpha1.GatewayConfiguration) *operatorv1alpha1.DataPlane {
+func (r *GatewayReconciler) provisionDataPlane(ctx context.Context,
+	log logr.Logger,
+	gateway *gatewayDecorator,
+	gatewayConfig *operatorv1alpha1.GatewayConfiguration) *operatorv1alpha1.DataPlane {
```
pmalek, line 211:
```suggestion
func (r *GatewayReconciler) provisionDataPlane(
	ctx context.Context,
	log logr.Logger,
	gateway *gatewayDecorator,
	gatewayConfig *operatorv1alpha1.GatewayConfiguration,
	) *operatorv1alpha1.DataPlane {
```

nit. Was that manually formatted or via a tool? `gofumpt`?

controllers/utils.go
```diff
@@ -48,8 +48,8 @@ type loggerShim struct {
 	logger logr.Logger
 }
 
-func (l loggerShim) Debug(msg string)   { l.logger.V(logging.InfoLevel).Info(msg) }
-func (l loggerShim) Info(msg string)    { l.logger.V(logging.InfoLevel).Info(msg) }
+func (l loggerShim) Debug(msg string)   { l.logger.V(logging.DebugLevel).Info(msg) }
+func (l loggerShim) Info(msg string)    { l.logger.V(logging.DebugLevel).Info(msg) }
```
pmalek, line 52: 🤔 Why do we need this?

controllers/utils.go
```diff
@@ -281,6 +279,18 @@ func debug(log logr.Logger, msg string, rawOBJ interface{}, keysAndValues ...int
 	}
 }
 
+func trace(log logr.Logger, msg string, rawOBJ interface{}, keysAndValues ...interface{}) {
+	if obj, ok := rawOBJ.(client.Object); ok {
+		kvs := append([]interface{}{"namespace", obj.GetNamespace(), "name", obj.GetName()}, keysAndValues...)
```
pmalek, line 284: I think that at some point we should write a common func that will deduplicate the keys in here to prevent logging `namespace` and `name` multiple times.

controllers/gateway_controller.go
mlavacca: manually formatted.

controllers/utils.go
mlavacca: because the cfssl logger logs `signed certificate with serial number 324652955489284542989245542923784676242674427411` with info severity every time a certificate is signed. I think we should convert it to the debug severity as this info is not meaningful when you just want info.

controllers/utils.go
mlavacca: yep, I agree, even if `keysAndValues` is an unused field at the moment.

pmalek (approved):

PR #282, opened 2022-09-01
test: each test has its own environment

### LINKED ISSUE #279: feat: enhance e2e test environment

### Problem Statement

As of now, all the e2e tests run in the same cluster and all the objects are reconciled by the same gateway operator instance. This leads to problems when there is the need to test in an isolated environment in which no tests have run before.

### Proposed Solution

_No response_

### Additional information

_No response_

### Acceptance Criteria

- [ ] each e2e test has a fresh and isolated environment in which to run

mlavacca:

Each e2e test creates its own environment and deploys the gateway operator. This means that all the e2e tests now run in isolation.

**Which issue this PR fixes**
Fixes #279 

**PR Readiness Checklist**:


shaneutt (approved):
Looks like this should do it. :+1:

test/e2e/webhook_test.go
```diff
@@ -13,11 +14,21 @@ import (
...
 )
 
 func TestDataplaneValidatingWebhook(t *testing.T) {
-	t.Log("start tests")
-	testNamespace, cleaner := setup(t)
+	var env environments.Environment
+	ctx, cancel := context.WithCancel(context.Background())
+	defer func() {
+		require.NoError(t, cleanupEnvironment(ctx, env))
+		cancel()
```
pmalek, line 25: Hmm if `require`'s assertions fails here then `cancel()` won't be called. This might be innocuous but ideally we'd prefer to always call `cancel()` to abort all operations that use it.

mlavacca: `cleanupEnvironment` needs a context, so I think it'd complain if we pass a closed context to it.

pmalek: Well that should work as well. I'd have to remind myself if in case of a failed assertion the `defer` is still executed but in general yes, something like that.

mlavacca: I added this here: https://github.com/Kong/gateway-operator/pull/269

shaneutt: Whoops good catch patryk: yes asserts only in `defer` blocks :+1:

PR #286, opened 2022-09-01
fix(ci): fix image testing workflow by using a lowercased github repository name

pmalek:

#264 introduced a new reusable workflow for building images and for testing them but apparently due to Kong's Github org starting with capital `K` we hit a snag with:

```
docker: invalid reference format: repository name must be lowercase.
```

https://github.com/Kong/gateway-operator/runs/8142185001?check_suite_focus=true

This PR fixes is my using `${{ github.repository }}` lowercased.

😒 

Example of what it looks in action: https://github.com/pmalek/gateway-operator/runs/8143823870?check_suite_focus=true#step:8:2


shaneutt (approved):
Thanks for linking the github discussion. You'd think there would be another way to handle this, but for now let's get this working and we can follow up.

Issue #272, opened 2022-08-31
DataPlane status should include addresses


shaneutt:
### Problem Statement

Currently a `DataPlane` resource has an owner relationship with a `Service` that provides addresses at which it can be reached, but we don't reflect this in the status. This makes it more difficult to use the `DataPlane` API independently from `Gateway`, and also makes `Watch()` semantics for the `Gateway` (which _does_ track it's own addresses in status) complicated. We should be able to rely on the `Owns(dataplane)` to ensure that `Gateway` addresses are eventually consistent (instead having to watch the `Services` directly as we don't want to start introducing several layers of indirection). We should add addresses to the status of the `DataPlane` and update them as the relevant `Services` change.

### Proposed Solution

- [ ] implement addresses in the `DataPlane` status, similar to (or perhaps even copying) `Gateway`'s implementation
- [ ] add tests that verify that both the `DataPlane` AND the `Gateway` are eventually consistent with their `Service` addresses
- [ ] remove the old temporary service watch code from the `Gateway`

### Additional information

Since I figured this was a bit of a bigger bit of work, I put a temporary workaround in the form of a service watch into the gateway controller. This should be removed in favor of `Owns()` doing all the work as a part of this effort.

### Acceptance Criteria

- [ ] as a user my `DataPlanes` include status to indicate the network addresses that they can be reached at
- [ ] as a user my `DataPlanes` and `Gateways` have address status that is eventually consistent with linked `Services` used to access the `DataPlane`

PR #270, opened 2022-08-31
fix: deployments on EKS

shaneutt:
**What this PR does / why we need it**:

This makes it possible to deploy the KGO and `Gateway` resources correctly on Amazon EKS clusters, which use hostnames for their `LoadBalancer` type `Service` addresses by default. It also adds sorting to prioritize `LoadBalancer` addresses for `Gateway` status addresses and enables `Service` watching for the `Gateway` controller so it can respond to updates to the `Service` created for a `DataPlane` by updating it's status addresses as needed.

**Which issue this PR fixes**

Fixes #259
Fixes #268 
Partially resolves #24 

**Special notes for your reviewer**:

Due to a [lack of EKS support in the KTF](https://github.com/Kong/kubernetes-testing-framework/issues/285) testing was done manually. Testing was done as such:

```console
eksctl create cluster --name kgo-test
eksctl utils write-kubeconfig --cluster kgo-test
make run &
k apply -f config/samples/gateway.yaml
```

Though we still want to add automated tests for this as a part of #24, I think ideally we should add this now without waiting for that so that we're able to demo on EKS sooner rather than later.

#272 was created as a follow up to this, as some of the watch logic added here was a good quick fix, but not ideal long term.

PR #300, opened 2022-09-05
Update gateway on service change

### LINKED ISSUE #192: Update Gateway status on Service change
Follow up from https://github.com/Kong/gateway-operator/pull/175#issuecomment-1221007406

Currently, the operator only performs a single initial pass to mark a Gateway ready and set addresses in status. If the state of its associated Service changes, the Gateway will retain its initial status. For example, starting with:

```
11:41:21-0700 esenin $ kubectl get svc
NAME                         TYPE           CLUSTER-IP      EXTERNAL-IP    PORT(S)                                     AGE
dataplane-kong-zb7bq-j25nl   LoadBalancer   10.96.101.238   192.168.16.0   80:31251/TCP,443:30890/TCP,8444:32191/TCP   5m38s
kubernetes                   ClusterIP      10.96.0.1       <none>         443/TCP                                     9m41s
```
and then changing the metallb address range results in a change to the Service IP, but not the corresponding Gateway IP:
```
11:43:21-0700 esenin $ kubectl get svc    
NAME                         TYPE           CLUSTER-IP      EXTERNAL-IP     PORT(S)                                     AGE
dataplane-kong-zb7bq-j25nl   LoadBalancer   10.96.101.238   192.168.200.0   80:31251/TCP,443:30890/TCP,8444:32191/TCP   7m46s
kubernetes                   ClusterIP      10.96.0.1       <none>          443/TCP                                     11m
11:43:33-0700 esenin $ kubectl get gateway
NAME   CLASS   ADDRESS        READY   AGE
kong   kong    192.168.16.0   True    9m46s
```
This presumably also extends to cases where a LoadBalancer loses its addresses entirely (not aware of a way to force metallb to do this offhand to confirm): if this occurs, the associated Gateway will keep its address and its ready state. We should instead update readiness and status on all Service-triggered reconciles:

- [ ] If a Gateway's Service's addresses change after they are initially set, the operator updates the Gateway status with the new addresses.
- [ ] If a Gateway's Service updates to have _no_ addresses, and the Gateway is currently marked Ready, the Gateway is marked not Ready.
- [ ] If a Gateway's Service updates to have addresses and the Gateway is currently not Ready, and has no other Conditions that would make it not Ready, the Gateway becomes Ready with the Service's addresses. This functionality is already in place, we just need to ensure we don't somehow break it.

There are some additional minor issues with the status handling that we should address while we're in the neighborhood:

- "Ready" is an overloaded term in function naming. There is a Ready Condition which can be either true or false, and it means "not Ready" when it is false. There's also the general English language definition of "ready", which means, well, "ready", and never means "not ready". Some functions use the first meaning (e.g. `SetReady` and `ensureGatewayMarkedReady` evaluate whether the controller should be marked ready, and sets either the true or false Ready Condition accordingly) and some use the second (`IsReady` returns true if the true Ready Condition is present), which is a bit confusing to read. We should clarify these names somehow.
- We set Ready conditions sometimes in the `ensureGatewayMarkedReady` helper (when we're setting a true Ready Condition) and sometimes in `Reconcile()` (when we're setting a false Ready Condition). IMO placing both updates in the same location would be better organized.

mlavacca:

**Which issue this PR fixes**
Fixes #192 

**PR Readiness Checklist**:


controllers/gateway_controller_reconciler_utils.go
```diff
@@ -95,6 +95,7 @@ func (r *GatewayReconciler) ensureGatewayConnectivityStatus(ctx context.Context,
 	}
 	svc := services[0]
 	if svc.Spec.ClusterIP == "" {
+		gateway.Status.Addresses = []gatewayv1alpha2.GatewayAddress{}
```
pmalek, line 98: ~~Do I get this right that we rely on the caller to issue an `Update` or `Patch` with this change?~~

~~I'm not sure I like this approach. I'd prefer this to be a bit more verbose and explicit~~

Hmm 🤔 That's the way it's been done for this whole `func`. Do we do this like that in every `ensure...()` func?

controllers/gateway_controller_test.go
```diff
@@ -0,0 +1,269 @@
...
+	networkPolicyName = "test-networkpolicy"
+
+	clusterIP      = "10.96.1.50"
+	loadBalancerIP = "172.18.1.18"
+)
+
+var (
+	dataplaneServiceName = fmt.Sprintf("svc-%s", dataplaneName)
+	IPAddressTypePointer = (*gatewayv1alpha2.AddressType)(pointer.StringPtr(string(gatewayv1alpha2.IPAddressType)))
+)
```
pmalek, line 44: Is there a reason this is outside of the test function? I'd prefer we not share this to prevent accidental changes.

controllers/gateway_controller_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+	gatewayClassName  = "test-gatewayclass"
+	gatewayName       = "test-gateway"
+	dataplaneName     = "test-dataplane"
+	controlplaneName  = "test-controlplane"
+	networkPolicyName = "test-networkpolicy"
+
+	clusterIP      = "10.96.1.50"
+	loadBalancerIP = "172.18.1.18"
+)
```
pmalek, line 39: The same for those

controllers/gateway_controller_reconciler_utils.go
mlavacca: Yes, `patchStatus` is called only from the `gateway_controller.go` file. No status is updated in this file. It has been reworked in the last couple of weeks to have consistency across our funcs.

controllers/gateway_controller_test.go
mlavacca: They are global because are addressed many times in this test (and probably in future tests). The alternative is having this hard coded; that is something I'd prefer to avoid.

controllers/gateway_controller_test.go
mlavacca: same as above.

controllers/gateway_controller_test.go
pmalek: I hear you but OTOH I treat "isolation" as one of the most important traits of good tests and by sharing variables like these we're breaking its promise. How would you feel about putting that in the test itself? This way whatever is needed for the test is right there. It also decreases the "code jumping" when reading through the test itself.

shaneutt:
In https://github.com/Kong/gateway-operator/pull/270 I added updates to the `Gateway` on changes of dataplane `Services`, specifically so that long-provisioning load-balancers (GKE in particular) would be eventually consistent. How does that change relate/interact with this?

shaneutt:
Also, https://github.com/Kong/gateway-operator/issues/272 I believe is the better long term solution.

mlavacca:
> In #270 I added updates to the `Gateway` on changes of dataplane `Services,` specifically so that long-provisioning load-balancers (GKE in particular) would be eventually consistent. How does that change relate/interact with this?

The only change in this PR is that the Gateway addresses are forced to be empty if the service has no `ClusterIP`  set. everything related to the eventual consistency of the Gateways shouldn't interfere.

